PR #7, opened 2020-04-14
Fix timer issue

yzgr2:
if the next timer is deleted by current timer's callback function,
the timer list is broken.


rzr:
please forward your patch upstream at:

https://iotivity.org/get-involved

WAvdBeek:
this problem is already solved on master

Issue #20, opened 2021-05-12
Blockwise timeout notify 503 not working properly for GET requests


WAvdBeek:
Blockwise timeout notify 503 not working properly for GET requests
For some reason I'm not able to fork and create a pull request (it is disabled), but I've found that there is an error in engine.c.

In blockwise operations, the mid is updated for each new block, thus the response_buffer's mid needs to be updated on each block. For GET operations, the response_buffer mid is updated correctly but the client_cb mid is not see here.

If there is a timeout (see here) oc_ri_free_client_cbs_by_mid is called. This function is supposed to notify with 503 the response handler. However, since cb->mid hasn't been updated with the new blocks mid, the transaction id is diferent and thus it doesn't notify the handler.

The solution would be to simply add the following: if(client_cb) client_cb->mid = response_mid; right after/before this.

If someone can explain how I can fork and do a pull request (maybe I just need permissions) I'll gladly do a pull request :)

DavidMarquezF:
I believe it can be closed now

PR #28, opened 2021-05-12
Fix blockwise response timeout

DavidMarquezF:
Fixes issue #20


WAvdBeek:
Hi David,
do you agree with the Contributor License Agreement?: Apache 2.0

DavidMarquezF:
> Hi David,
> do you agree with the Contributor License Agreement?: Apache 2.0

Yes

PR #42, opened 2021-05-31
api: fix oc_delayed_delete_resource function

Danielius1922:
Do not invoke oc_cloud_delete_resource directly
in oc_server_api.c, because not all targets link
with oc_cloud source code. Instead assign
oc_cloud_delete_resource to global pointer if
cloud is enabled and invoke this pointer (if set).

### DIFF
--- a/include/oc_api.h
+++ b/include/oc_api.h
@@ -1153,24 +1153,6 @@ bool oc_add_resource(oc_resource_t *resource);
 bool oc_delete_resource(oc_resource_t *resource);
 
 /**
- * @brief Callback invoked on resource before it is deleted by
- * oc_delayed_delete_resource.
- *
- * @param resource Resource to be deleted
- */
-typedef void (*oc_delete_resource_cb_t)(oc_resource_t *resource);
-
-/**
- * Sets the callback that gets invoked by oc_delayed_delete_resource
- * before each resource is deleted.
- *
- * @param callback The callback to set or NULL to unset it. If the function
- *                 is invoked a second time, then the previously set callback is
- *                 simply replaced.
- */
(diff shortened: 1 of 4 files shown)


Danielius1922:
@ondrejtomcik @jkralik

ondrejtomcik (approved):

ondrejtomcik (approved):

include/oc_api.h
```diff
@@ -1152,6 +1152,24 @@ bool oc_add_resource(oc_resource_t *resource);
...
+
+/**
+ * Sets the callback that gets invoked by oc_delayed_delete_resource
+ * before each resource is deleted.
+ *
+ * @param callback The callback to set or NULL to unset it. If the function
+ *                 is invoked a second time, then the previously set callback is
+ *                 simply replaced.
+ */
+void oc_set_on_delayed_delete_resource_cb(oc_delete_resource_cb_t callback);
```
jkralik, line 1171: I think this call belows to internal API. Because if user use it `cloud_init` override it.

WAvdBeek: if this is only for cloud, then I think this should be encapsulated with ifdefs

Danielius1922: > if this is only for cloud, then I think this should be encapsulated with ifdefs

Yeah, currently invoked only from cloud_init, so it can be ifdef-ed. It will be 4 ifdefs (whole header, include, static variable and invocation) and ifdefs worsen readability. Do we think it's worth it?

WAvdBeek: yes it is worth it...
we like to keep the code (without client) as small as possible)

jkralik (approved):

WAvdBeek (approved):

PR #47, opened 2021-06-03
Fix: While retrying to connect to the cloud, keep last error to CLOUD_OK

### LINKED ISSUE #35: Regression in test 1.4.5
This is regression with testing with CTT
commits 66faa4ab - eb95cf48 caused issue in test 1.4.5 Cloud: Last Error Code
Please see the logs:
CTT_Logs_Archive__3_.zip
It's in parts:

"Test IUT has to access non-existing cloud" and "Test create account return 4.01"

It could be reproduced every time.

FYI @mkettrick @Wouter

Hello,
before this change the cps was set to failed right after the first failure https://gitlab.iotivity.org/iotivity/iotivity-lite/-/blob/137c690a4d73d7979f5f12b7f1edbe42ee52c13d/api/cloud/oc_cloud_manager.c#L220 but the provisioning didn't failed until the retry count is over, therefore it has been updated to report registering status until the retry count is zero https://gitlab.iotivity.org/iotivity/iotivity-lite/-/blob/master/api/cloud/oc_cloud_manager.c#L216
This is described in OCF Device To Cloud Service Specification v 2.1.0 6.2.3.2.3  ... Device shall transition to the "failed" state, unless the Device autonomously re-attempts the registration ...
So this means that the test must be updated and not the implementation, right?
FYI @ondrejtomcik @jkralik

Hi All.
please use tag 2.2.3 of Iotivity of testing.
note that that master has already progressed, hence seeing this failure.
it still needs to be fixed for a future iotivity/ctt release (on the CTT side).

I've checked and tag 2.2.3 pass all of the testcases.
Which branch, tag, shall we use for the Night builds for regression testing?
When I asked this question last time I get answer that we should use master

Hi Marek,
yes, the master should be used for regression testing.
so this is a new CTT issue due to code changes after releasing 2.2.3.
Note that gitlab will be retired soon.
please start using https://github.com/iotivity/iotivity-lite master
note this github repo is now currently synced from gitlab.

Oxymoron79:
While retrying to connect to the cloud, keep last error to `CLOUD_OK`
This should fix #35 

@jkralik @WAvdBeek @ondrejtomcik

### DIFF
--- a/api/cloud/oc_cloud_manager.c
+++ b/api/cloud/oc_cloud_manager.c
@@ -213,7 +213,8 @@ error:
     cloud_set_cps_and_last_error(ctx, OC_CPS_FAILED, CLOUD_ERROR_RESPONSE);
   } else {
     ctx->store.status |= OC_CLOUD_FAILURE;
-    cloud_set_cps(ctx, OC_CPS_REGISTERING);
+    // While retrying, keep last error (clec) to CLOUD_OK
+    cloud_set_cps_and_last_error(ctx, OC_CPS_REGISTERING, CLOUD_OK);
   }
   return -1;
 }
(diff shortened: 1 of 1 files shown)


jkralik (approved):

jkralik:
Pls, could you review and merge it @WAvdBeek ? Thx

PR #48, opened 2021-06-07
Persist cloud store after cps and last_error were reset.

Oxymoron79:
It could happen, that after a `OC_RESET` the cloud store does not reset the values for `cps` (i.e. the reset value for `cps` is only in memory). That can cause the cloud manager to enter a wrong state, when the iotivity-lite server is restarted.

This PR fixes the issue.

Pleaes review: @jkralik


jkralik (approved):

jkralik:
@WAvdBeek could you merge it ? Thx

WAvdBeek (approved):

PR #56, opened 2021-06-22
linux: don't set SO_REUSEADDR for random port

jkralik:
system always peak free port and when SO_REUSEADDR is set, then multiple
process of device with small chance system can peak same port.
eg for 1000 process it is cca 20ports - replicated via docker.

### DIFF
--- a/port/linux/ipadapter.c
+++ b/port/linux/ipadapter.c
@@ -1315,11 +1315,6 @@ connectivity_ipv4_init(ip_context_t *dev)
     OC_ERR("setting pktinfo IPv4 option %d\n", errno);
     return -1;
   }
-  if (setsockopt(dev->server4_sock, SOL_SOCKET, SO_REUSEADDR, &on,
-                 sizeof(on)) == -1) {
-    OC_ERR("setting reuseaddr option %d", errno);
-    return -1;
-  }
   if (bind(dev->server4_sock, (struct sockaddr *)&dev->server4,
            sizeof(dev->server4)) == -1) {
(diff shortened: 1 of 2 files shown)


jkralik:

port/linux/ipadapter.c
```diff
@@ -1344,11 +1339,6 @@ connectivity_ipv4_init(ip_context_t *dev)
     OC_ERR("setting pktinfo IPv4 option %d\n", errno);
     return -1;
   }
-  if (setsockopt(dev->mcast4_sock, SOL_SOCKET, SO_REUSEADDR, &on, sizeof(on)) ==
```
WAvdBeek, line 1347: are you sure that the multicast socket is not fixed?

jkralik: No. this a bug - multicast socket must use `SO_REUSEADDR`. - Fixed

Danielius1922 (approved):

WAvdBeek:
Hi Josef.

just to be sure: did you run 2 devices on the same PC?
(you already confirmed that a single device is running, but this change is about reusing ports)

jkralik:
> 
> 
> Hi Josef.
> 
> just to be sure: did you run 2 devices on the same PC?
> (you already confirmed that a single device is running, but this change is about reusing ports)

I tested it with docker 1000 devices on the same PC. Without the patch I was able discover cca 980 devices because some devices shared ports.

WAvdBeek (approved):

ondrejtomcik (approved):

PR #64, opened 2021-07-24
Obt python

khaefner:
This is the first PR for the OBT web UI.


khaefner:
I'm going to delete this PR and replace it with a new one...I've done so many changes externally to make merging very difficult.

PR #65, opened 2021-07-27
Test App for TV Apps and Action Pattern

marktrayer:


WAvdBeek:
Hi Mark,

is there a change to the makefile?

PR #71, opened 2021-08-12
fix: during the wait for closing TLS connection, drop commands

jkralik:
When factory reset occurs, the device waits(2s) for closing TLS
connections. And during these 2s someone starts owning the device
again, he fails because the device closes all connections including
owning connection. For that, we drop all commands which come for
that time.

FYI: @ondrejtomcik, @Danielius1922

### DIFF
--- a/api/oc_main.c
+++ b/api/oc_main.c
@@ -64,6 +64,13 @@
 
 #include "oc_main.h"
 
+#ifdef OC_DYNAMIC_ALLOCATION
+#include <stdlib.h>
+static bool *drop_commands;
+#else  /* OC_DYNAMIC_ALLOCATION */
+static bool drop_commands[OC_MAX_NUM_DEVICES];
+#endif /* !OC_DYNAMIC_ALLOCATION */
+
 static bool initialized = false;
 static const oc_handler_t *app_callbacks;
 static oc_factory_presets_t factory_presets;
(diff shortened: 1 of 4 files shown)


api/oc_main.c
```diff
@@ -225,6 +232,12 @@ oc_main_init(const oc_handler_t *handler)
     oc_shutdown_all_devices();
     goto err;
   }
+#ifdef OC_DYNAMIC_ALLOCATION
+  drop_commands = (bool *)calloc(1, sizeof(bool));
```
Danielius1922, line 236: Is allocating memory just for one bool sufficient, shouldn't it be

```
calloc(OC_MAX_NUM_DEVICES, sizeof(bool));
```

?

jkralik: Fixed

ondrejtomcik (approved):

Danielius1922 (approved):

api/oc_main.h
```diff
@@ -39,6 +39,9 @@ typedef struct oc_random_pin_t
 
 bool oc_main_initialized(void);
 
+void oc_set_drop_commands(size_t device, bool drop);
```
WAvdBeek, line 42: please document what these new functions are doing

jkralik: Added.

WAvdBeek (approved):

marktrayer (approved):

PR #38, opened 2021-05-25
Jakarta

WAvdBeek:
Jakarta features

PR #72, opened 2021-08-13
api/cloud: include short tokens in deregister query

Danielius1922:
Some older cloud implementations expect the deregister
requests to have access token in the query otherwise
they regard the request as invalid. To support such
implementations we append the token to query if
the overall size of the request header is shorter
or equal to COAP_MAX_HEADER_SIZE.

### DIFF
--- a/api/cloud/oc_cloud_apis.c
+++ b/api/cloud/oc_cloud_apis.c
@@ -267,10 +267,29 @@ cloud_deregister(cloud_api_param_t *p)
 
   OC_DBG("try deregister");
   bool cannotConnect = true;
+  // This value is calculated by coap_oscore_serialize_message for deregister
+  // message with empty query parameters. The value should remain the same
+  // unless some global options are added to coap requests.
+  // The deregister request won't be sent if the total size of its header is
+  // greater than COAP_MAX_HEADER_SIZE, so we must ensure that the query
+  // is not too large.
+  // Some older cloud implementations require tokens in deregister requests.
+  // To facilitate support for such implementations we append access token
+  // to the request query if the resulting query size is within the limit.
+  #define DEREGISTER_EMPTY_QUERY_HEADER_SIZE 38
+  oc_string_t query = cloud_access_deregister_query(oc_string(ctx->store.uid),
+    oc_string(ctx->store.access_token), /*device*/0);
+  size_t query_size = oc_string_len(query);
+  oc_free_string(&query);
+  const char* access_token = NULL;
+  if (DEREGISTER_EMPTY_QUERY_HEADER_SIZE + query_size <= COAP_MAX_HEADER_SIZE) {
+    access_token = oc_string(ctx->store.access_token);
+  }
   if (oc_string(ctx->store.ci_server) && conv_cloud_endpoint(ctx) == 0 &&
       cloud_access_deregister(ctx->cloud_ep,
                               oc_string(ctx->store.uid),
-                              0,
+                              access_token,
+                              /*device*/0,
                               cloud_deregistered_internal,
                               p)) {
     cannotConnect = false;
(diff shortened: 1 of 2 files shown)


api/cloud/oc_cloud_apis.c
```diff
@@ -267,10 +267,29 @@ cloud_deregister(cloud_api_param_t *p)
...
+  const char* access_token = NULL;
+  if (DEREGISTER_EMPTY_QUERY_HEADER_SIZE + query_size <= COAP_MAX_HEADER_SIZE) {
+    access_token = oc_string(ctx->store.access_token);
+  }
   if (oc_string(ctx->store.ci_server) && conv_cloud_endpoint(ctx) == 0 &&
       cloud_access_deregister(ctx->cloud_ep,
                               oc_string(ctx->store.uid),
-                              0,
+                              access_token,
+                              /*device*/0,
```
jkralik, line 292: why there is 0 instead of the device?

api/cloud/oc_cloud_apis.c
```diff
@@ -472,8 +491,38 @@ cloud_access_register(oc_endpoint_t *endpoint, const char *auth_provider,
   return oc_do_post();
 }
 
+oc_string_t
+cloud_access_deregister_query(const char *uid, const char *access_token, size_t device)
+{
+  oc_string_t q_uid;
```
jkralik, line 497: pls initialization to zero value

api/cloud/oc_cloud_apis.c
```diff
@@ -472,8 +491,38 @@ cloud_access_register(oc_endpoint_t *endpoint, const char *auth_provider,
...
+  oc_string_t q_di;
+  oc_concat_strings(&q_di, "&di=", uuid);
+  oc_string_t q_uid_di;
+  oc_concat_strings(&q_uid_di, oc_string(q_uid), oc_string(q_di));
+  oc_free_string(&q_uid);
+  oc_free_string(&q_di);
+
+  oc_string_t q_uid_di_at;
+  if (access_token != NULL) {
+    oc_string_t q_at;
```
jkralik, line 511: pls initialization to zero value

api/cloud/oc_cloud_apis.c
```diff
@@ -472,8 +491,38 @@ cloud_access_register(oc_endpoint_t *endpoint, const char *auth_provider,
...
+  char uuid[OC_UUID_LEN] = { 0 };
+  oc_uuid_to_str(oc_core_get_device_id(device), uuid, OC_UUID_LEN);
+  oc_string_t q_di;
+  oc_concat_strings(&q_di, "&di=", uuid);
+  oc_string_t q_uid_di;
+  oc_concat_strings(&q_uid_di, oc_string(q_uid), oc_string(q_di));
+  oc_free_string(&q_uid);
+  oc_free_string(&q_di);
+
+  oc_string_t q_uid_di_at;
```
jkralik, line 509: pls initialization to zero value

api/cloud/oc_cloud_apis.c
Danielius1922: Not really sure, but looking in my original commit ( https://github.com/iotivity/iotivity-lite/pull/38/commits/c91cba16ae488be908d0217cd667e5a3f36649f6 ) it seems that the value for device has been always zero.

api/cloud/oc_cloud_apis.c
Danielius1922: I don't think it's necessary, `oc_concat_strings` creates a new `oc_string_t` from the second and third argument and saves it the first argument, essentially it's an initializer.

api/cloud/oc_cloud_apis.c
jkralik: could you change it to the device?

jkralik (commented):
pls just use the device instead of 0 and @WAvdBeek as a reviewer.

api/cloud/oc_cloud_apis.c
Danielius1922: Which value shoud I use though, ctx->store.device?

jkralik: yes

WAvdBeek (approved):

PR #92, opened 2021-09-01
Add CMake Buildsystem

tiniuclx:
This pull request:

- adds a CMake build system capable of building Iotivity-Lite on UNIX and Windows (tested using the Visual Studio project generator)
- adds `format` target, which will run `clang-format` locally & apply the changes to your working directory

Iotivity is set up to be built on all platforms, but only the Windows & Linux ports are built by CMake. Ports for other platforms must provide their own `iotivity-port` CMake target that the rest of the code links against.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -33,4 +33,5 @@ target_link_libraries(simpleclient
 )
 file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/simpleclient_creds)
 
+# include file used to compile the Device Builder application
 include(device-builder.cmake OPTIONAL)
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -24,7 +24,7 @@ if(UNIX)
         ${iotivity-lite_SOURCE_DIR}/api/oc_introspection.c
     )
     target_link_libraries(simpleclient
-        iotivity-client
+        iotivity-server-client
     )
     file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/simpleclient_creds)
 
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -9,6 +9,7 @@ target_link_libraries(simpleserver
     pthread
     m
 )
+file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/simpleserver_creds)
 
 add_executable(simpleserver_pki
     ${PROJECT_SOURCE_DIR}/simpleserver_pki.c
@@ -19,6 +20,7 @@ target_link_libraries(simpleserver_pki
     pthread
     m
 )
+file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/simpleserver_pki_creds)
 
 add_executable(simpleclient
     ${PROJECT_SOURCE_DIR}/simpleclient.c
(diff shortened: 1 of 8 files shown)


WAvdBeek (approved):

apps/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+    )
+    file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/simpleclient_creds)
+endif()
+
+# copy credentials, used by example applications.
+add_custom_target(copy-pki-certs ALL
+    COMMAND ${CMAKE_COMMAND} -E copy_directory
+    ${PROJECT_SOURCE_DIR}/pki_certs
+    ${PROJECT_BINARY_DIR}/pki_certs
+)
```
CiaranWoodward, line 66: I don't think this does as intended. This adds the copy operation to the 'all' target so that when you type for instance ``make`` or ``make all``, this operation is executed. But if you type ``make simpleclient`` or any other specific target, it will not be.

The solution to this is one of:
- Just copy the entire directory at configure time (I think this makes the most sense and is simplest)
  - use ``execute_process`` for this
- Add this target as a dependency of every application target or the pki library (not recommended, custom targets are always out of date)
- use ``add_custom_command(TARGET <target_here> POST_BUILD COMMAND <command here>`` to run the command every time a certain target is built (such as one of the core libraries)

onboarding_tool/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+    add_executable(onboarding_tool
+        ${PROJECT_SOURCE_DIR}/obtmain.c
+        ${iotivity-lite_SOURCE_DIR}/api/oc_introspection.c
+    )
+
+    target_link_libraries(onboarding_tool
+        iotivity-client
+    )
+
+    file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/onboarding_tool_creds)
```
CiaranWoodward, line 13: It seems odd that this is required everywhere, does the application not make its own directories?

onboarding_tool/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+    # copy credentials
+    add_custom_target(copy-pki-certs-obt ALL
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
+        ${iotivity-lite-apps_SOURCE_DIR}/pki_certs
+        ${PROJECT_BINARY_DIR}/pki_certs
+    )
```
CiaranWoodward, line 20: Same here

CiaranWoodward (changes requested):
Looks great! Hopefully will be much nicer to maintain and extend than the makefiles. I imagine github-actions support to make sure it continues to work is coming in future?

Couple of comments inline, but nothing that is fundamentally wrong if we need to merge it quickly.

onboarding_tool/CMakeLists.txt
CiaranWoodward: That seems very odd, as you then can't run the application from a different folder or move it, but I guess this solves the problem for now 👍

apps/CMakeLists.txt
tiniuclx: The problem is that we want to make sure that the certs in the build dir update whenever the certs in the source dir update. I am not sure that `execute_process` accomplishes this. I don't really think we lose anything by copying the certs more frequently than strictly necessary. Thoughts?

PR #105, opened 2021-09-07
Workaround to avoid infinite loop when no query parms are present in a

marktrayer:
retrieve to /oic/res


SiMet:
I check this branch with the CTT and I'm able to discover Iotivity.
Looks like your changes fixed the issue.

tiniuclx (approved):

PR #126, opened 2021-09-29
Add resource remotecontrol to server_certification_test

SiMet:
Add resource oic.r.remotecontrol to sample app to add resource with query parameter action into regression testing


SiMet:
~format

PR #131, opened 2021-10-04
Add support for if.startup and if.startup.revert in server_certification_test

SiMet:
Add support of feature to have possibility to test in during the Night builds


WAvdBeek (approved):

WAvdBeek:
~format

PR #137, opened 2021-10-20
seperate check if wellknown core is secure or not.

WAvdBeek:


WAvdBeek:
I was on a call, hence did not have time to fix that error just yet, so thanks for doing that marek!

PR #138, opened 2021-10-20
Perform discovery within cloud proxy using /oic/sec/doxm?deviceuuid=...

tiniuclx:
Since this pull request relies on an implementation of CR3426, the relevant additions are disabled by default & are only in effect when `OC_DOXM_UUID_FILTER` is defined within `cloud_proxy.c`.

The benefit of this approach is that significantly less traffic is generated by the multicast requests, which improves reliability on low bandwidth networks such as Thread.

Tested using an OCF-over-Thread end to end system using PLGD & a prototype implementation of CR3426. From the point of view of the cloud, the behaviour is identical: the device is discovered & proxied as normal.

This PR also creates a CMake target for the cloud proxy & corrects some misleading API documentation.

### DIFF
--- a/apps/cloud_proxy.c
+++ b/apps/cloud_proxy.c
@@ -165,7 +165,7 @@
 /* perform discovery using /oic/sec/doxm, which generates significantly less
  * traffic when compared to /oic/res discovery
  */
-#define OC_DOXM_UUID_FILTER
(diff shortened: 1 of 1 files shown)


tiniuclx:
@WAvdBeek Should be ready for the merge now

PR #153, opened 2021-11-08
allow set OC_MIN_APP_DATA_SIZE for dynamic allocation

jkralik:
encoding response buffer doesn't alloc full OC_MAC_APP_DATA_SIZE,
but it starts at OC_MIN_APP_DATA_SIZE, and if during encoding
needs more memory, it calls realloc up to OC_MAC_APP_DATA_SIZE

formula:
buffer_size = 2*(buffer_size + needed)


jkralik:
@tiniuclx 
> My biggest concern with this pull request is that it sometimes increases peak memory consumption. Let's say you start off with a buffer of 6kB & you find out that it must be reallocated. The next valid size up is OC_MAX_APP_DATA_SIZE - 7kB in our case. If the heap does not have any unallocated bytes past our 6kB buffer, the allocator will have to allocate an entirely new 7kB buffer, copy all the data across & free it afterwards. This briefly allocates a total of 13kB which exceeds the peak memory allocation of 7kB that we had before this pull request.

The main purpose is to minimalize the size of the response and drop `OC_MAX_OBSERVE_SIZE` , but yes I understand your concerns. I can change the strategy that if `new size > OC_MAX_APP_DATA_SIZE/4` then I will allocate a full buffer. The worst-case will be `OC_MAX_APP_DATA_SIZE/4 + OC_MAX_APP_DATA_SIZE`.
 
> If you want to avoid this issue, a strategy of allocating the maximum OC_MAX_APP_DATA_SIZE and reallocating downwards (making the big buffer smaller as you learn about its correct size) would work.

If you call for eg:
```c
oc_set_min_app_data_size(8*1024);
oc_set_max_app_data_size(8*1024);
```
The behavior will same as before. The reallocation will be not called. But I will add this improvement that when I know size then will downsize the buffer.

> 
> The other thing that we need to be aware of is that `realloc` can (and does!) fail at any time, including when you are right in the middle of building up an `oc_rep_t`. If the error code is not checked after every operation, we can run into a situation such as not noticing a realloc failure while you are building an outgoing `oc_rep_t`, which leads to the following macro writing to invalid memory.

Realloc is used only for encoding to cbor. It has nothing to do with parsing (oc_rep_t is used only for that). 
And failing is covered by line https://github.com/iotivity/iotivity-lite/blob/4767e573b33bdc9e4876fd5a7713f918aafa44c9/api/oc_rep.c#L102

tiniuclx:
> The main purpose is to minimalize the size of the response and drop `OC_MAX_OBSERVE_SIZE` , but yes I understand your concerns. I can change the strategy that if `new size > OC_MAX_APP_DATA_SIZE/4` then I will allocate a full buffer. The worst-case will be `OC_MAX_APP_DATA_SIZE/4 + OC_MAX_APP_DATA_SIZE`.
> 
This is still a significant increase in peak dynamic memory consumption - for some of our embedded targets `OC_MAX_APP_DATA_SIZE/4 + OC_MAX_APP_DATA_SIZE` is 20% of our available heap!

The core of the problem is that you need a large buffer for observing `/oic/res`, while we need the observe buffer to be as small as possible so that multiple resources can be observed without running out of memory & still being able to respond when other requests come through.

I don't understand why this problem must be solved at runtime (with increased memory consumption) when we know about the problem right now & can configure the stack appropriately. I would be happy to set `OC_MAX_OBSERVE_SIZE` to 7k - I can even make the pull request. This would let you perform observes on `/oic/res` without affecting our embedded deployments as we are free to change its value within the port.

jkralik:
> > The main purpose is to minimalize the size of the response and drop `OC_MAX_OBSERVE_SIZE` , but yes I understand your concerns. I can change the strategy that if `new size > OC_MAX_APP_DATA_SIZE/4` then I will allocate a full buffer. The worst-case will be `OC_MAX_APP_DATA_SIZE/4 + OC_MAX_APP_DATA_SIZE`.
> 
> This is still a significant increase in peak dynamic memory consumption - for some of our embedded targets `OC_MAX_APP_DATA_SIZE/4 + OC_MAX_APP_DATA_SIZE` is 20% of our available heap!

I will describe the implementations why I want to use my approach.

General setup
- 10KB free memory
- OC_MAX_APP_DATA_SIZE = 7KB
- 1 response with length 6KB is sending via blockwise 
- processing one request and encoding response

1. current master branch: 7KB is consumed by BWT, and for encoding response, there is not enough memory, because it always allocates 7KB, so response never comes
2. your suggestion(just downsizing buffer to length response): 6KB is still consumed by BWT and for encoding response, there is not enough memory to allocate 7KB even the device wants to send a response with 1KB.
3. this PR: if 6KB is consumed (i will add downsizing) by BWT, the device can send a response up to 1.4KB length. Still you can use your approach set `oc_set_min_app_data_size` and `oc_set_max_app_data_size` with same value and you will get same behaviour as point 2.


> 
> The core of the problem is that you need a large buffer for observing `/oic/res`, while we need the observe buffer to be as small as possible so that multiple resources can be observed without running out of memory & still being able to respond when other requests come through.
> 

Observation of oic/res will be optional and driven by a macro `OC_DISCOVERY_RESOURCE_OBSERVABLE` so if the device doesn't have much memory then just comment it in oc_config.h

> I don't understand why this problem must be solved at runtime (with increased memory consumption) when we know about the problem right now & can configure the stack appropriately. I would be happy to set `OC_MAX_OBSERVE_SIZE` to 7k - I can even make the pull request. This would let you perform observes on `/oic/res` without affecting our embedded deployments as we are free to change its value within the port.

`OC_MAX_OBSERVE_SIZE` from my point of view hit macro is weird. Because it just solve observation partially but what about any commands PUT/GET/POST/RETRIEVE? This is why I try to solve the issue globally with this PR. Could you check if this PR works with your chip? If not then yes we need to have it ...

ondrejtomcik:
I also don't like current solution where you set some limits during the compile time. Jozef described it well and solution he prepared IMHO effectively solves the problem and even improves the mem consumption for constrained devices by not allocating mem of a size which is not even used. It's much more clean solution.

WAvdBeek:
> I also don't like current solution where you set some limits during the compile time. Jozef described it well and solution he prepared IMHO effectively solves the problem and even improves the mem consumption for constrained devices by not allocating mem of a size which is not even used. It's much more clean solution.

Hi All,
I do not like the approach taken by Josef. Period!.
Such change will make sure that IoTivity does not work anymore on small devices.

ondrejtomcik:
Constructive feedback and good discussion @WAvdBeek. Thanks

WAvdBeek:
> Constructive feedback and good discussion @WAvdBeek. Thanks

one should think and try to engage before starting to make changes.

ondrejtomcik:
We are not complaining or emotional. MR is a place for collaboration. Not a place for emotions 🙂

ondrejtomcik:
@tiniuclx could you please look into the update Jozef made? Would such a proposal work for you? If not, what about having a call about it next week?

Thanks and have a nice weekend

security/oc_store.c
```diff
@@ -29,16 +29,19 @@
...
 #include <stdlib.h>
 #else /* OC_DYNAMIC_ALLOCATION */
 #define OC_APP_DATA_STORAGE_BUFFER
+#ifndef OC_APP_DATA_BUFFER_SIZE
+#define OC_APP_DATA_BUFFER_SIZE OC_MAX_APP_DATA_SIZE
+#endif /* OC_APP_DATA_BUFFER_SIZE */
 #endif /* !OC_DYNAMIC_ALLOCATION */
 
+#ifdef OC_APP_DATA_STORAGE_BUFFER
+static uint8_t buf[OC_APP_DATA_BUFFER_SIZE];
```
Danielius1922, line 42: `buf` seems like a dangerous name for a file-static variable. 

I guess it's a deliberate shadowing with the malloc-ed buf when `OC_APP_DATA_STORAGE_BUFFER`, '-Wshadow' is enabled thus the compiler should catch shadowing problems, so I guess it's alright.

tiniuclx:
I've reviewed the code as it is right now, and it looks good. However, @jkralik mentioned on the call that he wants to do a rebase to fix the merge conflict, as well as add a couple of macros. @WAvdBeek  I would be happy to see this merged as soon as that happens.

api/oc_rep.c
```diff
@@ -38,10 +119,30 @@ void
...
 }
 
+#ifdef OC_DYNAMIC_ALLOCATION
+void
+oc_rep_new_realloc(uint8_t **out_payload, int size)
+{
+  g_err = CborNoError;
+  g_enable_realloc = true;
+  g_buf_size = size;
+  g_buf = (uint8_t **)out_payload;
```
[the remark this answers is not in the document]
jkralik: fixed

security/oc_store.c
jkralik: added check  `-Wshadow` to build

api/oc_blockwise.c
```diff
@@ -111,6 +117,17 @@ oc_blockwise_free_buffer(oc_list_t list, struct oc_memb *pool,
   oc_memb_free(pool, buffer);
 }
 
+static uint32_t
+oc_blockwise_get_buffer_size(oc_blockwise_state_t *buffer)
+{
+#ifdef OC_DYNAMIC_ALLOCATION
+  return buffer->buffer_size;
```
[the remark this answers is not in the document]
jkralik: fixed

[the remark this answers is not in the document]
jkralik: `oc_memb_alloc` - calls internaly `calloc` or `memset`

jkralik:
> I've reviewed the code as it is right now, and it looks good. However, @jkralik mentioned on the call that he wants to do a rebase to fix the merge conflict, as well as add a couple of macros. @WAvdBeek I would be happy to see this merged as soon as that happens.

@tiniuclx @WAvdBeek
Done 😄 
- I added macro `OC_REP_ENCODING_REALLOC` which is not set in default - it enables reallocation during encoding the representation to cbor.
- `OC_MAX_OBSERVE_SIZE` is still a valid macro and you can set it for your device, be default is not set
- I added 2 devices to test which verify `OC_REP_ENCODING_REALLOC`

ondrejtomcik (approved):

api/oc_mnt.c
```diff
@@ -62,11 +62,21 @@ post_mnt(oc_request_t *request, oc_interface_mask_t iface_mask, void *data)
...
+                       OC_MAX_APP_DATA_SIZE);
+#else  /* OC_DYNAMIC_ALLOCATION && OC_REP_ENCODING_REALLOC */
     oc_rep_new(request->response->response_buffer->buffer,
-               OC_MAX_APP_DATA_SIZE);
+               request->response->response_buffer->buffer_size);
+#endif /* !OC_DYNAMIC_ALLOCATION || !OC_REP_ENCODING_REALLOC */
     oc_rep_start_root_object();
     oc_rep_set_boolean(root, fr, false);
     oc_rep_end_root_object();
+#ifdef OC_DYNAMIC_ALLOCATION
```
Danielius1922, line 76: Shouldn't this have the same `ifdef` as above?

```defined(OC_DYNAMIC_ALLOCATION) && defined(OC_REP_ENCODING_REALLOC)```

Comment (Danielius1922) [reply api/oc_mnt.c:76]: Actually, doesn't need either `oc_rep_shrink_encoder_buf` checks that both are defined and does nothing if they aren't.

File: api/oc_ri.c [hunk 5]
```diff
@@ -1066,21 +1064,36 @@ oc_ri_invoke_coap_entity_handler(void *request, void *response, uint8_t *buffer,
 /* Alloc response_state. It also affects request_obj.response.
  */
 #ifdef OC_BLOCK_WISE
+#ifdef OC_DYNAMIC_ALLOCATION
+#ifdef OC_REP_ENCODING_REALLOC
+  bool response_state_allocated = false;
+#endif /* OC_REP_ENCODING_REALLOC */
+  bool REP_ENCODING_REALLOC_created = false;
```
Danielius1922, line 1071: :see_no_evil:

jkralik: fixed 😄

api/oc_mnt.c
jkralik: We want to use shrink buffer for DYNAMIC memory so I changed the implementation.
OC_REP_ENCODING_REALLOC - just allows you to change OC_MIN_APP_DATA_SIZE:

- if OC_REP_ENCODING_REALLOC is set, the default value of OC_MIN_APP_DATA_SIZE is 256
- if OC_REP_ENCODING_REALLOC  is not set, the default value of OC_MIN_APP_DATA_SIZE is OC_MAX_APP_DATA_SIZE

Danielius1922: Sounds good to me.

Danielius1922 (approved):

tiniuclx (approved):

WAvdBeek (approved):

PR #161, opened 2021-11-23
fix cancel observation via BWT

jkralik:
When the client is observing a resource via BWT, then the client
cancel it, observation destroys cancelation response which is not
expected by
https://datatracker.ietf.org/doc/html/rfc7641#section-3.6 .

### DIFF
--- a/messaging/coap/observe.c
+++ b/messaging/coap/observe.c
@@ -245,7 +245,11 @@ coap_remove_observer(coap_observer_t *o)
     oc_string(o->resource->uri) + 1, oc_string_len(o->resource->uri) - 1,
     &o->endpoint, OC_GET, query, (query) ? strlen(query) : 0,
     OC_BLOCKWISE_SERVER);
-  if (response_state) {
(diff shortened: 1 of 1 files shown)


ondrejtomcik (approved):

messaging/coap/observe.c
```diff
@@ -245,7 +245,11 @@ coap_remove_observer(coap_observer_t *o)
     oc_string(o->resource->uri) + 1, oc_string_len(o->resource->uri) - 1,
     &o->endpoint, OC_GET, query, (query) ? strlen(query) : 0,
     OC_BLOCKWISE_SERVER);
-  if (response_state) {
+  // If response_state->payload_size == 0 it means, that this blockwise state
+  // doesn't below to the observer. Because the observer always sets
```
Danielius1922, line 249: typo: below -> belong?

Danielius1922 (approved):

messaging/coap/observe.c
jkralik: fixed

WAvdBeek (approved):

PR #163, opened 2021-12-02
Enable thread-sanitizer builds on Linux

Danielius1922:


Danielius1922:
An occasional segfault occurs when we try onboarding multiple devices in parallel, from the generated core dump and the subsequent analysis the issue seemed like a data race. Thread sanitizer is a data race detector available for C/C++ in clang and GCC, so enabling it for analysis seems prudent (additional there are also memory, undefined behavior and leak sanitizers available).




After running the test case with thread-sanitized build several data races were detected. Here is the [log](https://github.com/iotivity/iotivity-lite/files/7645616/cloud_server.tsan.log). The stack trace with the `refresh_endpoints_list` matches the segfault from core dumps. I'm working on fixing that, however there are also other reports. It would be useful to run thread-sanitized build with all available test cases to see if it generates additional reports.

WAvdBeek (approved):

PR #164, opened 2021-12-03
util: atomic operations on Linux and Windows

Danielius1922:


jkralik (approved):

port/linux/ipadapter.c
```diff
@@ -706,9 +717,18 @@ process_interface_change_event(void)
...
     for (i = 0; i < num_devices; i++) {
       ip_context_t *dev = get_ip_context_for_device(i);
-      oc_network_event_handler_mutex_lock();
-      refresh_endpoints_list(dev);
-      oc_network_event_handler_mutex_unlock();
+      if (dev == NULL) {
+        continue;
+      }
+      bool swapped = false;
+      int expected = OC_ATOMIC_LOAD(dev->flags);
```
jkralik, line 724: why you don't call `OC_ATOMIC_LOAD32` because `int` is 32bit ?

port/linux/ipcontext.h
```diff
@@ -96,6 +102,7 @@ typedef struct ip_context_t
   pthread_mutex_t rfds_mutex;
   fd_set rfds;
   int shutdown_pipe[2];
+  OC_ATOMIC int flags;
```
jkralik, line 105: why you don't use the `int32_t` type?

port/linux/ipadapter.c
Danielius1922: GCC's `__atomic_` functions work for (correctly aligned) integers of all sizes, so it's not necessary to use integers of specific size. But I can use `int32_t` and `OC_ATOMIC_LOAD32` if you think it would help with clarity.

jkralik: Yes. It is more clear.
Maybe we don't need a similar function as `OC_ATOMIC_LOAD` without suffix 32. Some ports don't support atomic 64bits operations.

Danielius1922:
In this PR I've fixed one of the issues (one with `refresh_endpoints_list` call in the stack traces) reported by TSAN ([tsan.8.log](https://github.com/iotivity/iotivity-lite/files/7660715/tsan.8.log)). However, I've fixed it only for Linux. I've checked code in other ports and they all have the same implementation, now maybe on some platforms `volatile` is atomic, however for some it definitely isn't (Windows, Android, ...). I don't want to port this blindly, are there any dedicated platform maintainers that could check it and port it if the issue also occurs on the given platform?

Additionally, TSAN reports issues in common code (in `util/oc_process.c` for example). I'm not sure how we want to handle it, I've created `oc_atomic.h` header which has implementation of atomics for Linux (on GCC) and Windows, it defaults to `volatile` on other platforms. I can fix the issues in common code using this library, but again verification on all platforms should be done.

jkralik (approved):

jkralik:
From my view, we can use `oc_atomic.h` in `util/oc_process.c`  to guard shared variables instead of using `volatile`.
@tiniuclx What do you think?

WAvdBeek (approved):

PR #171, opened 2021-12-15
Collection: fix segfault on create

Danielius1922:


messaging/coap/observe.c
```diff
@@ -674,8 +694,6 @@ coap_notify_collections(oc_resource_t *resource)
   }
 #endif /* OC_DYNAMIC_ALLOCATION */
 
-  int num_links = 0;
```
Danielius1922, line 677: Maybe this was supposed to be increased in the for cycle that follows this code and iterates over collections with links? However, exists in master currently, it is never modified.

jkralik (approved):

PR #173, opened 2021-12-16
Fix data races

Danielius1922:


.github/workflows/plgd-device-test.yml
```diff
@@ -21,23 +21,42 @@ jobs:
...
-            tag: cloud-server-discovery-resource-observable-debug:debug
-
+            file: docker/apps/Dockerfile.cloud-server-debug
+            tag: dbg
+            args: "CLOUD=1 SECURE=1 CREATE=1 MNT=1 DEBUG=1 OICRES_OBSERVABLE=1 OSCORE=0 ASAN=0 TSAN=0"
+          - name: cloud-server-discovery-resource-observable-debug-asan
+            file: docker/apps/Dockerfile.cloud-server-debug
+            tag: dbg
+            args: "CLOUD=1 SECURE=1 CREATE=1 MNT=1 DEBUG=1 OICRES_OBSERVABLE=1 OSCORE=0 ASAN=1 TSAN=0"
+          - name: cloud-server-discovery-resource-observable-debug
```
jkralik, line 42: cloud-server-discovery-resource-observable-debug-tsan

api/oc_core_res.c
```diff
@@ -296,7 +299,7 @@ oc_core_con_handler_post(oc_request_t *request, oc_interface_mask_t iface_mask,
 size_t
 oc_core_get_num_devices(void)
 {
-  return device_count;
+  return g_device_count;
```
jkralik, line 302: `OC_ATOMIC_LOAD32(g_device_count);`

.github/workflows/plgd-device-test.yml
```diff
@@ -21,23 +21,42 @@ jobs:
...
+            tag: dbg
+            args: "CLOUD=1 SECURE=1 CREATE=1 MNT=1 DEBUG=1 OSCORE=0 ASAN=0 TSAN=1"
           - name: cloud-server-discovery-resource-observable-debug
-            file: docker/apps/Dockerfile.cloud-server-discovery-resource-observable-debug
-            tag: cloud-server-discovery-resource-observable-debug:debug
-
+            file: docker/apps/Dockerfile.cloud-server-debug
+            tag: dbg
+            args: "CLOUD=1 SECURE=1 CREATE=1 MNT=1 DEBUG=1 OICRES_OBSERVABLE=1 OSCORE=0 ASAN=0 TSAN=0"
+          - name: cloud-server-discovery-resource-observable-debug-asan
```
ondrejtomcik, line 38: We had an agreement that at the moment we introduce more combinations, we won't use names but refactor it to tags as it's really messy.

ondrejtomcik (approved):

.github/workflows/plgd-device-test.yml
Danielius1922: Fixed.

api/oc_core_res.c
Danielius1922: Fixed.

jkralik (approved):

Danielius1922:
I've updated the two plgd testing github workflows. It runs plgd tests with cloud_server device compiled with ASAN and TSAN enabled, if a problem is found it crashes and prints a report. I've found one heap-use-after-free and several data races by running this.

api/oc_core_res.c
```diff
@@ -47,7 +49,7 @@ static oc_platform_info_t oc_platform_info;
 
 static bool announce_con_res = false;
 static int res_latency = 0;
-static size_t device_count = 0;
+static OC_ATOMIC_UINT32_T g_device_count = 0;
```
WAvdBeek, line 52: why change the type?

Danielius1922: `size_t` is one of those annoying types whose size depends on the architecture and for atomic operations I must have known width before choosing the operation (at least on Windows). 

Changing the type requires the least amount of changes. There are other options (using `sizeof` to detect the true width of the variable; or not using atomic operations, but protecting the variable with a mutex.), but this seemed the best. Though I kept `size_t` as return value of `oc_core_get_num_devices`, so that's now inconsistent. That shouldn't be a problem, unless we support platforms where `size_t` is 16-bit, which would mean that the return value of `oc_core_get_num_devices` would get truncated for values larger than UINT16_T max (65535). If that's a relevant case for us then I suggest adding a check to `oc_core_add_new_device` so the value is never greater than SIZE_MAX.

WAvdBeek: well, I do not think the approach of the atomic is the right one anyway, please use critical sections around the size_t variables.

jkralik: @WAvdBeek 
Using critical sections is more expensive than atomic operations for CPU and generally, atomic operations are used for counters like device_count.  Please, could you explain it more?

WAvdBeek (approved):

PR #178, opened 2022-01-28
CMake: add support for clang-tidy static analysis

Danielius1922:


Danielius1922:
I've copied the CMake code and workflow from my repo I'm working on. Works like a charm but the output is too noisy right now (see the "static-analysis" job output). I guess I can disable the analysis of headers and some checks.

@jkralik Is it OK that CMake uses C11 to compile everything but in the port/linux Makefile we use gnu99? It's inconsistent.

PR #181, opened 2022-02-02
Extend CMake build

Oxymoron79:
Extend / refactor the CMake build to imitate the Makefile build:

- Build dynamic and static libraries for client, server and client-server:
  - Create Object libraries to assemble the dynamic and static libraries from.
  - The sources from the `port/*` folders need to be built separately for client and server object libraries, due to a `OC_CLIENT` compilation switch in the source code.
  -  Build patched mbedtls from source (as the Makefile does) instead of using the mbedtls CMake project. This made the creation of static and dynamic libraries easier.
- Extend build configuration variables (CACHED)
- Generate pkg-config files using the template files `port/linux/*.pc.in`
- Generate CMake package containing all library variants `{client,server,client-server}-{static,shared}` as components in the `iotivity-lite` namespace. This provides a modern use of the library in CMake, e.g:
   ```
   find_package(iotivity-lite REQUIRED)
   if(USE_STATIC_LIB)
       target_link_libraries(${PROJECT_NAME} PRIVATE iotivity-lite::client-server-static)
   else(USE_STATIC_LIB)
       target_link_libraries(${PROJECT_NAME} PRIVATE iotivity-lite::client-server-shared)
   ```
- Add installation steps with `COMPONENTS` to allow future packaging with [CPack](https://cmake.org/cmake/help/latest/module/CPack.html).
- Build unit tests using the [CTest](https://cmake.org/cmake/help/latest/module/CTest.html) module.
- Build binaries from the `tests` directory with a custom `check` target that executes the `tools/check.py` script.
- Build most example applications (can be disabled by setting the build variable `BUILD_EXAMPLE_APPLICATIONS` to `OFF`). The example applications are currently not installed, because the use a relative path for the storage directory that requires them to be executed in a specific directory, which a installation might change.
- Build the python client library only as dynamic library (as in Makefile). It is also not installed (as in Makefile).

I have manually tested the CMake build on Ubuntu 18.04 with CMake 3.10.2.

### TODO

- [x] The unittests `apitest`, `platformtest` and `messagingtest` fail if `OC_IPV4` is not set. Is this also the case with the Makefile build?
     --> No, with the Makefile build the unittests pass. There was a compile definition missing for mbedtls, Fixed in https://github.com/iotivity/iotivity-lite/pull/181/commits/633c82ae1d1cdcda422901a6a708ca41832042ba
- [x] The test `client_get_linux_test` hangs during execution. Is this also the case with the Makefile build?
      --> Yes, with the Makefile build the test hangs as well.
- [x] @Danielius1922 : Could the PR https://github.com/iotivity/iotivity-lite/pull/178 be integrated here?
- [x] Windows builds with MSVC and MinGW should be possible, but I haven't tested it yet. Help with that would be welcome!


deps/mbedtls.cmake
```diff
@@ -0,0 +1,31 @@
...
+    set(OC_REAPPLY_MBEDTLS_PATCHES OFF CACHE BOOL
+        "By default, mbedTLS patches are applied upon the first CMake Configure. Set this to ON to reapply the patches on the next configure."
+         FORCE
+    )
+endif()
+
+file(GLOB MBEDTLS_SRC
+    ${PROJECT_SOURCE_DIR}/deps/mbedtls/library/[a-l]*.c
+    ${PROJECT_SOURCE_DIR}/deps/mbedtls/library/md*.c
+    ${PROJECT_SOURCE_DIR}/deps/mbedtls/library/[n-x]*.c
```
Danielius1922, line 14: As I'm looking at the list of files in `deps/mbedtls/library` isn't this skipping only the `memory_buffer_alloc.c` file? Seems simpler to glob `*.c` here and then add the file to `list(REMOVE_ITEM MBEDTLS_SRC)` call below.

CMakeLists.txt
```diff
@@ -24,213 +14,559 @@ set(OC_IDD_API_ENABLED ON CACHE BOOL "Enable the Introspection Device Data API."
...
 set(OC_REPRESENTATION_REALLOC_ENCODING_ENABLED OFF CACHE BOOL "Enable realloc during encoding the representation.")
+set(OC_COLLECTIONS_IF_CREATE_ENABLED OFF CACHE BOOL "Enable RT factory for collections.")
+set(OC_MNT_ENABLED OFF CACHE BOOL "Enable maintenance resource.")
+set(OC_SOFTWARE_UPDATE_ENABLED OFF CACHE BOOL "Enable software update resource.")
+set(OC_WKCORE_ENABLED OFF CACHE BOOL "Enable well-known core resource.")
+set(OC_OSCORE_ENABLED OFF CACHE BOOL "Enable oscore support.")
+set(OC_IPV4_ENABLED OFF CACHE BOOL "Enable IPv4 support.")
+set(OC_DNS_LOOKUP_IPV6_ENABLED OFF CACHE BOOL "Enable IPv6 DNS lookup.")
+
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
Danielius1922, line 25: [nitpick] I usually turn this only for dynamic library targets, but I don't know if there are any significant optimizations skipped in static executables/libraries because of this.

CMakeLists.txt
```diff
@@ -24,213 +14,559 @@ set(OC_IDD_API_ENABLED ON CACHE BOOL "Enable the Introspection Device Data API."
...
+if(OC_DISCOVERY_RESOURCE_OBSERVABLE_ENABLED)
+    list(APPEND PRIVATE_COMPILE_DEFINITIONS "OC_DISCOVERY_RESOURCE_OBSERVABLE")
+endif()
+
+if(OC_REPRESENTATION_REALLOC_ENCODING_ENABLED)
+    list(APPEND PRIVATE_COMPILE_DEFINITIONS "OC_REP_ENCODING_REALLOC")
+endif()
+
+if(OC_COLLECTIONS_IF_CREATE_ENABLED)
+    list(APPEND PUBLIC_COMPILE_DEFINITIONS "OC_COLLECTIONS_IF_CREATE")
```
Danielius1922, line 69: I think the general collections enabler define `OC_COLLECTIONS` is missing.

CMakeLists.txt
```diff
@@ -24,213 +14,559 @@ set(OC_IDD_API_ENABLED ON CACHE BOOL "Enable the Introspection Device Data API."
...
+endif()
+list(APPEND SERVER_SRC ${PORT_SRC})
+list(APPEND CLIENT_SRC ${PORT_SRC})
+
+if(OC_SECURITY_ENABLED)
+    file(GLOB SECURITY_SRC
+        ${PROJECT_SOURCE_DIR}/security/oc_*.c
+    )
+    list(REMOVE_ITEM SECURITY_SRC
+        ${PROJECT_SOURCE_DIR}/security/oc_obt*.c
```
Danielius1922, line 135: I don't think `list` operations support globbing. You can first `file(GLOB)` the obt files and then remove them.

WAvdBeek:
can we do these changes in stages?
e.g. add first the tests
then add obt & java builds etc.
it is hard to follow all the changes in one go.

deps/mbedtls.cmake
Oxymoron79: You're right. Fixed in https://github.com/iotivity/iotivity-lite/pull/181/commits/7501f8b0c88911708bd389345ea2635c9dff63b5

CMakeLists.txt
Oxymoron79: You're right. Fixed in https://github.com/iotivity/iotivity-lite/pull/181/commits/6cafe455fddb893be043fa64cd3acdf873fda661

Danielius1922 (approved):

Danielius1922:
Good job, nice work!

Oxymoron79:
> can we do these changes in stages? e.g. add first the tests then add obt & java builds etc. it is hard to follow all the changes in one go.

Yes, the diff on the main `CMakeLists.txt` file is quite extensive. This is because it is basically a complete rewrite needed to compile static and dynamic libraries at the same time.

I tried to have a clean commit log to make it easier to understand. 

- Up to commit https://github.com/iotivity/iotivity-lite/pull/181/commits/bc17c61c0ccff91545449c49ae477ca82aad6c20 deals only with the compilation of the libraries and unit tests. IMO that would be the first stage for a pull request. But even here the diff on the main `CMakeLists.txt` file is already quite extensive. From here on, the commits are quite self-contained.
- The commit https://github.com/iotivity/iotivity-lite/pull/181/commits/953ea7b368606f9b05a8203a91e116011cb5cca3 adds the onboarding tool. 
- The commit https://github.com/iotivity/iotivity-lite/pull/181/commits/31b14740de00abbf742749ee193966b78929f581 extends the example applications.
- The commit https://github.com/iotivity/iotivity-lite/pull/181/commits/1a9bcd1d1e79533d0eacb0d2f41d977612ad87de adds the test applications
- The commit https://github.com/iotivity/iotivity-lite/pull/181/commits/41157ed20dd0f30734bb23fd24d1a86aac871522 adds the python client dynamic library

The SWIG/Java build is not in this PR.

I can create separate PRs from these commits if you prefer, but I'm not sure that it reduces the diff on the main `CMakeLists.txt` file.

Oxymoron79:
> Good job, nice work!

Thanks, I'm glad you like it!

CMakeLists.txt
Oxymoron79: I came across this question as well, but then I saw that the Makefile also uses PIC to compile the objects. So I switched it on globally as this matches the Makefile build.

As a future improvement we can compile the object libraries twice - once with PIC and once without.

Oxymoron79:
During testing of the library built with MSVC I came across a issue when trying to link against the DLL: Since the library doesn't export any symbols, it is missing the import library that is required by the MSVC toolchain to link against a DLL. It's the same issue a reported in issue https://github.com/iotivity/iotivity-lite/issues/182.

It works fine though when the library is built with the MinGW Toolchain.

So until the symbols of the library are exported, I would suggest to disable the build of a DLL with the MSVC toolchain. Would you be OK with that for now?

Danielius1922:
> It works fine though when the library is built with the MinGW Toolchain.
That's probably down to the fact that it behaves as `GCC` or `clang` on Unix, by default all symbols have "default" visibility, which means that all are exported. However, usually exporting all symbols in shared lib is not something you want, so the general praxis on Unix is to change the default visibility to "hidden" and only mark the public API functions with a macro to be exported. I think this could be done when solving issue #182, so both Unix and Windows export similar set of functions.

ondrejtomcik (approved):

WAvdBeek (approved):

PR #185, opened 2022-02-09
CMake updates

Danielius1922:


Danielius1922:
@Oxymoron79 I have iotivity-lite as a submodule, I've called `add_subdirectory` with the main iotivity CMakeLists.txt and I've been getting this error all morning:

> Target "client-static" INTERFACE_INCLUDE_DIRECTORIES property contains
path:
>
>  "/home/danielius/GIT/device/iotivity-lite/"
>
> which is prefixed in the source directory.

I've finally found the problem. It seems that target_include_directories prepends the absolute path of the project directory to generator expresion when it contains a ";" but it's not quoted. This results in the `INTERFACE_INCLUDE_DIRECTORIES` property having unexpected values with absolute paths. So either the generator expression with ";" has to be quoted (`"$<BUILD_INTERFACE:${MBEDTLS_INCLUDE_DIR}>"`) or split into separate items without the separator.

Danielius1922:
@Oxymoron79, this is OT, but maybe you'll able to help me here. I'm trying to use `CPack` to create `.tgz` package with my binary, `libiotivity-lite-client.so`, `libiotivity-lite-client-server.so` and the corresponding `.pc` files. When I use `CPack` without any additional settings and call `make package` then I get an archive that contains every library (both `.a` and `.so`) and all `.pc` files. However, I only want the two shared libs and their corresponding headers and `.pc` files. I can do it manually in my CMake, but that seems fragile. I guess the better solution would be to create stand-alone install components for each shared library? Or any better ideas?

Oxymoron79:
> So either the generator expression with ";" has to be quoted ("$<BUILD_INTERFACE:${MBEDTLS_INCLUDE_DIR}>") or split into separate items without the separator.

If I remember correctly, I had issues with the CMake package or the example applications when I added absolute paths (using the `${PROJECT_SOURCE_DIR}` variable) to the BUILD_INTERFACE generator expression - but I didn't try quoting the whole generator expression.

If that works, I would prefer to quote the generator expression, because this allows the use of List variables.

Danielius1922:
> > So either the generator expression with ";" has to be quoted ("$<BUILD_INTERFACE:${MBEDTLS_INCLUDE_DIR}>") or split into separate items without the separator.
> 
> If I remember correctly, I had issues with the CMake package or the example applications when I added absolute paths (using the `${PROJECT_SOURCE_DIR}` variable) to the BUILD_INTERFACE generator expression - but I didn't try quoting the whole generator expression.
> 
> If that works, I would prefer to quote the generator expression, because this allows the use of List variables.

Everything compiles and links with this change, so it should be alright.

This is the `INCLUDE_DIRECTORIES` property value of `client-static` target for current the master version:
```
/home/danielius/GIT/iotivity-lite/$<BUILD_INTERFACE:.;/home/danielius/GIT/iotivity-lite/include;/home/danielius/GIT/iotivity-lite/port;/home/danielius/GIT/iotivity-lite/port/linux>;$<INSTALL_INTERFACE:include/iotivity-lite>;/home/danielius/GIT/iotivity-lite/$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite;/home/danielius/GIT/iotivity-lite/include;/home/danielius/GIT/iotivity-lite/port/linux;/home/danielius/GIT/iotivity-lite/deps/mbedtls/include>
```

and the value from this PR:
```
$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite>;$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite/include>;$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite/port>;$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite/port/linux>;$<INSTALL_INTERFACE:include/iotivity-lite>;$<BUILD_INTERFACE:/home/danielius/GIT/iotivity-lite;/home/danielius/GIT/iotivity-lite/include;/home/danielius/GIT/iotivity-lite/port/linux;/home/danielius/GIT/iotivity-lite/deps/mbedtls/include>
```

In the master version the for installation `/home/danielius/GIT/iotivity-lite/` is added (`BUILD_INTERFACE` expression hides the "." and other paths, but not the prefix), hence I got the warning about the `/home/danielius/GIT/iotivity-lite/` interface include.

Danielius1922:
> ...
>
> So I would prefer to not package the `.pc` files together with the `.so` files.
>
> ...

Yeah, makes sense, thank you. 

We can separate those, I guess that means 2 components per shared library (eg. for libclient-shared.so -> client-shared component with the .so files and client-shared-dev with the rest of the files). I think that the only really annoying part will be getting the correct list of public headers for each stand-alone shared lib.

Oxymoron79:
> > ...
> > So I would prefer to not package the `.pc` files together with the `.so` files.
> > ...
> 
> Yeah, makes sense, thank you.
> 
> We can separate those, I guess that means 2 components per shared library (eg. for libclient-shared.so -> client-shared component with the .so files and client-shared-dev with the rest of the files). I think that the only really annoying part will be getting the correct list of public headers for each stand-alone shared lib.

I would propose one component per shared library (libiotivity-lite-client.so -> iotivity-lite-client package, libiotivity-lite-server.so -> iotivity-lite-server package, libiotivity-lite-client-server.so -> iotivity-lite-client-server package) and one single iotivity-lite-dev package containing all the header files, `.pc` files the CMake package module and the static version of all library variants.

This way a developer can switch between the variants without needing to install a separate development package. And if he decides to use the shared library, the application can depend on one of the shared library packages to minimize the installation footprint on the target.  
This also makes it easier for us to to define the components (i.e. the current CMake already has the big `dev` installation component prepared, just the `bin` component needs to be changed).

What do you think of this proposal?

Oxymoron79 (approved):

jkralik (approved):

ondrejtomcik (approved):

Issue #182, opened 2022-02-02
Export functions through DLL for other libraries to use


LordMike:
I'm not entirely sure how this is meant to work, but if I want to interop with the API from C#, I had to do the following:

* Compile as a dll (please provide binaries :))
* Add `__declspec(dllexport)` to functions, to ensure they're exported

Could these things be done in the project itself, so that I and others can interop with the produced binary?

Oxymoron79:
Yes, that can be done in the project, but in a portable way. This is often achieved with a pre-processor macro like this one: https://github.com/zeromq/libzmq/blob/master/include/zmq.h#L62

LordMike:
Yes, exactly. I had a hard time adding these locally, as there are _so_ many methods, and I couldn't quickly identify which ones should be exported. It would have been easier if all public methods had been defined using a macro, so I could modify that and thereby add the export declaration.

A macro could be helpful also to ensure a consistent API surface, such as the `oc_` prefix.

PR #186, opened 2022-02-11
Test shared library on Windows

Danielius1922:


Danielius1922:
This is POC for solving issue #182. I've created a small shared library (uuid-shared) and test application (uuid-app), uuid-app links with uuid-shared and it compiles successfully on both Linux and Windows.
If the naming and location is fine then we can merge the `oc_export.h` header. Though going through all header files and adding the annotation will take a while. Maybe we can do it by parts somehow?

WAvdBeek (approved):

WAvdBeek:
> This is POC for solving issue #182. I've created a small shared library (uuid-shared) and test application (uuid-app), uuid-app links with uuid-shared and it compiles successfully on both Linux and Windows. If the naming and location are fine then we can merge the `oc_export.h` header. Though going through all header files and adding the annotation will take a while. Maybe we can do it by parts somehow?

Note that from python we are just using the oc_python.c file... so if we have an header file for that code, then one can build C# like the python code. C# has a similar interface as python c_types, e.g. add the call format to the dll in the C# code.
https://www.nag.com/IndustryArticles/Calling_C_Library_DLLs_from_C_Sharp.pdf

Danielius1922:
> > This is POC for solving issue #182. I've created a small shared library (uuid-shared) and test application (uuid-app), uuid-app links with uuid-shared and it compiles successfully on both Linux and Windows. If the naming and location are fine then we can merge the `oc_export.h` header. Though going through all header files and adding the annotation will take a while. Maybe we can do it by parts somehow?
> 
> Note that from python we are just using the oc_python.c file... so if we have an header file for that code, then one can build C# like the python code. C# has a similar interface as python c_types, e.g. add the call format to the dll in the C# code. https://www.nag.com/IndustryArticles/Calling_C_Library_DLLs_from_C_Sharp.pdf

You mean adding `oc_python.h` with the `OC_API` annotation for non-static functions or am I misunderstanding?

Danielius1922:
> correct. then we can also run the python (client) code on windows. not sure if we want to do the same API as JAVA, that is a considerable amount of work.
Sure, no problem, I can do it, shouldn't take that long to do the Windows part.

include/oc_export.h
```diff
@@ -0,0 +1,42 @@
...
+
+#if defined _WIN32 || defined __CYGWIN__ || defined __MINGW32__
+#define OC_EXPORT __declspec(dllexport)
+#define OC_IMPORT __declspec(dllimport)
+#elif __GNUC__ >= 4
+#define OC_EXPORT __attribute__((visibility("default")))
+#define OC_IMPORT
+#else /* __GNUC__ < 4 */
+#warning "Shared libraries not supported"
+#define OC_EXPORT __attribute__((visibility("default")))
```
Danielius1922, line 28: #define OC_EXPORT

CMakeLists.txt
```diff
@@ -405,8 +405,33 @@ if(NOT MSVC)
...
+)
+set_target_properties(uuid-shared PROPERTIES
+    OUTPUT_NAME "iotivity-lite-uuid"
+    VERSION ${PROJECT_VERSION}
+    SOVERSION ${PROJECT_VERSION_MAJOR}
+)
+
+# Python client
+if(UNIX OR WIN32)
     set(client-python-lib-obj
```
Danielius1922, line 435: Hmm, I don't think I can reuse the .o files. I need different annotations when I compile for `static` or `shared` lib.

include/oc_export.h
```diff
@@ -0,0 +1,42 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+#ifndef OC_EXPORT_H
+#define OC_EXPORT_H
+
+#if defined _WIN32 || defined __CYGWIN__ || defined __MINGW32__
```
Oxymoron79, line 20: I'm not sure whether `__CYGWIN__` and `__MINGW32__` use the `__declspec()` keyword. If I remember correctly, they also use use the `__attribute__` keyword. Is that correct?

include/oc_export.h
```diff
@@ -0,0 +1,42 @@
...
+#define OC_EXPORT __attribute__((visibility("default")))
+#define OC_IMPORT
+#else /* __GNUC__ < 4 */
+#warning "Shared libraries not supported"
+#define OC_EXPORT
+#define OC_IMPORT
+#endif /* defined _WIN32 || defined __CYGWIN__ || defined __MINGW32__ */
+
+#ifdef OC_SHARED_LIBRARY
+#ifdef OC_SHARED_LIBRARY_EXPORTS
```
Oxymoron79, line 33: Nitpick:
```suggestion
#ifdef OC_SHARED_LIBRARY_EXPORT
```
In my opinion this would be more consistent with `OC_EXPORT`.

include/oc_export.h
Danielius1922: Seems that for `__CYGWIN__` and `__MINGW32__` _both_ ways work. So I can simplify the expression and remove the `__CYGWIN__` and `__MINGW32__` subconditions. (Similar as it is in https://github.com/zeromq/libzmq/blob/master/include/zmq.h#L62 .)

include/oc_export.h
Danielius1922: Yeah, I agree. By default CMake adds `${TARGET_NAME}Exports` compile definition to shared library targets, so I kept the plural when I redefined it, but there is no reason for it.

CMakeLists.txt
Danielius1922: @Oxymoron79 I guess the solution here is to create shared version of the object libraries (`common-obj` and `common-obj-shared`, etc.), since they have differing compile definitions. Though they differ only in the compile definitions, so the CMake code for both targets will be very similar. I should probably create a CMake function that creates a library (both static and shared) target to avoid most of the duplicity.

Oxymoron79: Is that really necessary? The point of the object libraries is to generate static and dynamic libraries without the need to recompile the source...

Is your reasoning for this the `declspec(dllexport)` annotation for Windows? If so, I doubt that different object files are needed, because the user of the DLL has to change the annotation to `declspec(dllimport)` when linking to the binary (which was compiled with `declspec(dllexport)`). But my knowledge of Windows development is limited - so take that with a grain of salt :wink:

include/oc_export.h
Oxymoron79: > Seems that for `__CYGWIN__` and `__MINGW32__` _both_ ways work.

Ah, interesting - I didn't know that.

CMakeLists.txt
Danielius1922: I would like avoid the duplicity if possible, but when I compile `libiotivity-lite-client-python.so.2.2.5` and examine the public symbols using `nm -D` command - I see that all global functions from `common-obj`, `tinycbor-master`, `client-obj` and `python-obj` are public. Thus, adding `C_VISIBILITY` to the final shared library target had no effect. It seems that I have to compile the `.o` files with the macro definitions for a shared library, linking is not enough. However, I haven't tried it the other way around, add definitions for shared library to the object library targets. If static library targets won't break then that could be the way.

(OT: iotivity builds by CYGWIN and MINGW seem to be broken, but I think I've found the issue; fun for Friday afternoon :D )

Oxymoron79: If I understand https://cmake.org/cmake/help/latest/policy/CMP0063.html correctly, you can set this policy to `NEW` and then set the `C_VISIBILITY` on the object library.

Danielius1922: Thanks, that looks indeed relevant. Though I still think that this code is problematic:

```C
#ifdef OC_SHARED_LIBRARY
#ifdef OC_SHARED_LIBRARY_EXPORT
#define OC_API OC_EXPORT
#else /* !OC_SHARED_LIBRARY_EXPORT*/
#define OC_API OC_IMPORT
#endif /* OC_SHARED_LIBRARY_EXPORT */
#else  /* !OC_SHARED_LIBRARY */
#define OC_API
#endif /* OC_SHARED_LIBRARY */
```

I'm reading about `dllexport` and if I understand it correctly, whatever the compiler does with the attribute is compiled into the `.o` file. I see no way around the fact that the `.o` files are different between static and shared libraries. Shared libraries need `OC_SHARED_LIBRARY` and `OC_SHARED_LIBRARY_EXPORT` definitions during compilation. The only way this works is by compiling the object files with those definitions and hoping that a static library doesn't have a problem with that. 
Though I can think of one side effect, if you compile a static lib from such object files and then link that static lib into a dynamic library, I think the symbols from the static lib will be exposed as global symbols in the final shared library files since they will have the export attribute.

Oxymoron79: > (OT: iotivity builds by CYGWIN and MINGW seem to be broken, but I think I've found the issue; fun for Friday afternoon :D )

I came across this issue as well. One of our Windows developers found that adding this:
```
#ifdef __MINGW32__
#undef interface
#endif
```
in `port/windows/tcpadapter.c` and `port/windows/network_addresses.c`  after the include statements fixes the build.

To me, this seems like quite a hack - so I didn't propose the fix upstream. And I also don't know whether CYGWIN and MINGW build are supported by iotivity-lite.

Oxymoron79: > Thanks, that looks indeed relevant. Though I still think that this code is problematic:
> 
> ```c
> #ifdef OC_SHARED_LIBRARY
> #ifdef OC_SHARED_LIBRARY_EXPORT
> #define OC_API OC_EXPORT
> #else /* !OC_SHARED_LIBRARY_EXPORT*/
> #define OC_API OC_IMPORT
> #endif /* OC_SHARED_LIBRARY_EXPORT */
> #else  /* !OC_SHARED_LIBRARY */
> #define OC_API
> #endif /* OC_SHARED_LIBRARY */
> ```
> 
> I'm reading about `dllexport` and if I understand it correctly, whatever the compiler does with the attribute is compiled into the `.o` file. I see no way around the fact that the `.o` files are different between static and shared libraries. Shared libraries need `OC_SHARED_LIBRARY` and `OC_SHARED_LIBRARY_EXPORT` definitions during compilation. The only way this works is by compiling the object files with those definitions and hoping that a static library doesn't have a problem with that.

I think that it is correct to apply the visibility also on static libraries (btw the current visibility `default` made all symbols globally visible anyways). So, applying the compile definitions to the object library would promote the correct symbol visibility to the static and the shared library, right?

> Though I can think of one side effect, if you compile a static lib from such object files and then link that static lib into a dynamic library, I think the symbols from the static lib will be exposed as global symbols in the final shared library files since they will have the export attribute.

Yes, that's true. I found this question (with an answer) also on stackoverflow: https://stackoverflow.com/questions/2222162/how-to-apply-fvisibility-option-to-symbols-in-static-libraries.

Danielius1922:
> > (OT: iotivity builds by CYGWIN and MINGW seem to be broken, but I think I've found the issue; fun for Friday afternoon :D )
> 
> I came across this issue as well. One of our Windows developers found that adding this:
> 
> ```
> #ifdef __MINGW32__
> #undef interface
> #endif
> ```
> 
> in `port/windows/tcpadapter.c` and `port/windows/network_addresses.c` after the include statements fixes the build.
> 
> To me, this seems like quite a hack - so I didn't propose the fix upstream. And I also don't know whether CYGWIN and MINGW build are supported by iotivity-lite.

Yes, I spent a good half an hour trying to decipher a weird compilation error on a line with a simple definition of a variable named interface. :D Redefining a common word like `interface` using a macro seems like a bad idea, but I guess it was necessary for some reason. I would just rename the variables (`intf` or `iface` or whatever).

I think at least MINGW was previously supported or intended to be supported, see the cmake-windows workflow:
https://github.com/iotivity/iotivity-lite/blob/88c84f5f1ab81705bcb24ebe6e7766234cdfd0db/.github/workflows/cmake-windows.yml#L25-L29

I had to do some additional updates to the CMakeLists.txt (`if(UNIX)` is true in both `CYGWIN` and `MINGW`), but after that I was able to make the compilation succeed on both platforms. I think we can get support for both for pretty cheap.

WAvdBeek:
> > > (OT: iotivity builds by CYGWIN and MINGW seem to be broken, but I think I've found the issue; fun for Friday afternoon :D )
> > 
> > 
> > I came across this issue as well. One of our Windows developers found that adding this:
> > ```
> > #ifdef __MINGW32__
> > #undef interface
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in `port/windows/tcpadapter.c` and `port/windows/network_addresses.c` after the include statements fixes the build.
> > To me, this seems like quite a hack - so I didn't propose the fix upstream. And I also don't know whether CYGWIN and MINGW build are supported by iotivity-lite.
> 
> Yes, I spent a good half an hour trying to decipher a weird compilation error on a line with a simple definition of a variable named interface. :D Redefining a common word like `interface` using a macro seems like a bad idea, but I guess it was necessary for some reason. I would just rename the variables (`intf` or `iface` or whatever).
> 
> I think at least MINGW was previously supported or intended to be supported, see the cmake-windows workflow:
> 
> https://github.com/iotivity/iotivity-lite/blob/88c84f5f1ab81705bcb24ebe6e7766234cdfd0db/.github/workflows/cmake-windows.yml#L25-L29
> 
> I had to do some additional updates to the CMakeLists.txt (`if(UNIX)` is true in both `CYGWIN` and `MINGW`), but after that I was able to make the compilation succeed on both platforms. I think we can get support for both for pretty cheap.

nope, this yaml file was just copied from somewhere else and I made it work for visual studio only.
e.g. as precursor of deprecating the visual studio file under port/windows. 
maybe the cygwin worked when using the linux port (although I never tested that myself).

Danielius1922:
> nope, this yaml file was just copied from somewhere else and I made it work for visual studio only. e.g. as precursor of deprecating the visual studio file under port/windows. maybe the cygwin worked when using the linux port (although I never tested that myself).

I see, we can get `Mingw` and `Cygwin` with some minor code and CMake changes. Though the CMake doesn't run tests, so successful compilation might not mean that everything is in order.

CMakeLists.txt
Danielius1922: Yeah, seems like we should be able to get both static and shared libraries from object on Unixes. Not sure what will happen on MSVC, but I guess if my testing app links and runs it should be alright ...?

Though if the definitions will be added to both static and shared libraries I should change the names: `OC_SHARED_LIBRARY` -> `OC_LIBRARY` and `OC_SHARED_LIBRARY_EXPORT` -> `OC_LIBRARY_EXPORT`

Oxymoron79: > Yeah, seems like we should be able to get both static and shared libraries from object on Unixes. Not sure what will happen on MSVC, but I guess if my testing app links and runs it should be alright ...?

Yes, I think so as well.

> Though if the definitions will be added to both static and shared libraries I should change the names: `OC_SHARED_LIBRARY` -> `OC_LIBRARY` and `OC_SHARED_LIBRARY_EXPORT` -> `OC_LIBRARY_EXPORT`.

Sounds good to me.

Danielius1922: It seems to work. Compilation of testing apps that use static lib (`uuid-app-static`) and shared lib (`uuid-app-shared`) generated from object lib succeeds on both Linux and Win (see PR jobs). I've also examined `uuid-shared` with `nm -D` on Linux and with `dumpbin /exports` on Win10 locally and the lib only has the manually exposed symbols visible. 

Additional notes:

- CMP0063 is set to NEW by default, so I skipped setting it explicitly
- When compiling a static lib into a shared lib, gcc (at least my 9.3 version) seems to automatically hide all symbols from static library. So one must add compiler flags to make them visible, however this is not an interesting use case currently, so I didn't pursue it further.

OK, so I'll remove the uuid-* testing stuff and we can start annotating public functions and enabling compilation of DLLs on Windows. @WAvdBeek which functions are considered to be public, only those with headers in `include/` folder? Any other? (TBH, in my app that links dynamically with libiotivity-client-server.so, I use functions from all over the place, from `api`, `security`, `port` and `util` folders at least. But I'll work it out.)

WAvdBeek (approved):

WAvdBeek:
what needs to be done before merge?

Danielius1922:
> what needs to be done before merge?

Well, I've created testing apps (`uuid-app`, `uuid-shared-app`) and libraries (`uuid-obj`, `uuid-static`, `uuid-shared`) and I don't think we should merge those into master. However, those are an example of how to do it, so we should create at least one library target with static and shared versions that work. I guess the `client` library is the smallest? We need to annotate all public functions with the `OC_API` macro, I can start doing it, but I'm not sure which functions exactly are considered public. Only those in the `include/` folder? Any others?

WAvdBeek:
> > what needs to be done before merge?
> 
> Well, I've created testing apps (`uuid-app`, `uuid-shared-app`) and libraries (`uuid-obj`, `uuid-static`, `uuid-shared`) and I don't think we should merge those into master. However, those are an example of how to do it, so we should create at least one library target with static and shared versions that work. I guess the `client` library is the smallest? We need to annotate all public functions with the `OC_API` macro, I can start doing it, but I'm not sure which functions exactly are considered public. Only those in the `include/` folder? Any others?

oc_python.h, so that we can use the python code on windows.  For me that would be enough for a merge.

WAvdBeek:
> > oc_python.h, so that we can use the python code on windows. For me that would be enough for a merge.
> 
> I've removed the testing apps and `iotivity-lite-client-python.dll` compiled on Windows / libiotivity-lite-client-python.so on Linux, though on Linux symbols from `common-obj`, `tinycbor-master`, `client-obj` and `mbedtls` are exported as part of the lib. I think I can hide all symbols from targets `mbedtls` and `tinycbor-master`, those are third-party libs, no reason to expose their symbols as part of any iotivity library. For targets `common-obj` and `client-obj` I think some functions should be exported.

not sure if it is an actual issue that those symbols are exported. those symbols are used internally in the shared object/dll anyway.

Danielius1922:
> not sure if it is an actual issue that those symbols are exported. those symbols are used internally in the shared object/dll anyway.

Yes, that's not a problem that the shared library uses those symbols internally, the issue is that the shared library exports those symbols (at least on Unixes). That is they are available to a binary that links with the shared library.
The default behavior is different on MSVC and on GCC-compatible compilers (gcc, clang, mingw, cygwin). MSVC hides symbols by default, so global function from those mentioned targets are not visible to the clients of the `.dll`. GCC on other hand exposes symbols by default, so all global functions from the mentioned targets are visible to the clients of the `.so`.

Take the `client-python-shared` which is now built by this CMake on both Windows and on Unix:
* for Windows (with MSVC): the `.dll` only exposes functions annotated in the `oc_python.h`
* for Unix: the `.so` exposes functions from `oc_python.h`, but also all global functions from `common-obj`, `tinycbor-master`, 
`client-obj`, `mbedtls`.

I think we should have same or at least very similar exposed API on all platforms, so additional work is needed. But `oc_python.c` stuff is workable now, so I suggest we merge this PR and the remaining work will be done later.

PR #190, opened 2022-02-14
Update github workflow for Windows

Danielius1922:
Virtual environment windows-latest now defaults to windows-2022. Update workflows to work with this new environment.


Danielius1922:
@WAvdBeek it seems that `windows-latest` now means `windows-2022` in GitHub actions.

WAvdBeek (approved):

PR #192, opened 2022-02-16
Fix cloud registration during initialization

Oxymoron79:
I noticed that during the initialization of the cloud manager it tries to register to a cloud ([cloud_register](https://github.com/iotivity/iotivity-lite/blob/master/api/cloud/oc_cloud_manager.c#L295) is called), even though there is no cloud configuration stored.

Eventually this leads to the `cps` property of the `/CoapCloudConfResURI` to change to `OC_CPS_FAILED`.

@jkralik proposed to initialize the `cis` property with NULL (instead of "coaps+tcp://127.0.0.1") to fix the issue, but that wasn't enough. I noticed that now `cloud_store_initialize` is basically a duplicate of `cloud_store_deinit`, so I removed `cloud_store_deinit` along the way.

To fix the issue, I added a check `ctx->store.cps == OC_CPS_READYTOREGISTER` in `cloud_start_process` to prevent the call of `cloud_register` during initialization. To make sure that this check does not pass erroneously, I set the `cps` to `OC_CPS_UNINITIALIZED` when the device was deregistered.

### DIFF
--- a/api/cloud/oc_cloud_manager.c
+++ b/api/cloud/oc_cloud_manager.c
@@ -28,6 +28,7 @@
 #include "util/oc_list.h"
 #include "util/oc_memb.h"
 #ifdef OC_SECURITY
+#include "security/oc_pstat.h"
 #include "security/oc_tls.h"
 #endif /* OC_SECURITY */
 #include <stdint.h>
@@ -122,6 +123,11 @@ cloud_start_process(oc_cloud_context_t *ctx)
 {
   ctx->retry_count = 0;
   ctx->retry_refresh_token_count = 0;
+
+  oc_sec_pstat_t *pstat = oc_sec_get_pstat(ctx->device);
+  if (pstat->s != OC_DOS_RFNOP && pstat->s != OC_DOS_RFPRO) {
+    return;
+  }
 
   if (ctx->store.status == OC_CLOUD_INITIALIZED &&
       ctx->store.cps == OC_CPS_READYTOREGISTER) {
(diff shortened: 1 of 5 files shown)


api/cloud/oc_cloud_manager.c
```diff
@@ -123,7 +123,8 @@ cloud_start_process(oc_cloud_context_t *ctx)
   ctx->retry_count = 0;
   ctx->retry_refresh_token_count = 0;
 
-  if (ctx->store.status == OC_CLOUD_INITIALIZED) {
+  if (ctx->store.status == OC_CLOUD_INITIALIZED &&
```
jkralik, line 126: maybe we can check also if the device is in the `RFNORM` state. What do you think ?
```
#include "security/oc_pstat.h"
...
 oc_sec_pstat_t *pstat = oc_sec_get_pstat(ctx->device);
 if (pstat->s != OC_DOS_RFNOP) {
   return;
 }
 ```

Comment (Oxymoron79) [reply api/cloud/oc_cloud_manager.c:126]: @jkralik Yes, I aggree. That's a good suggestion. I will try it out.

Comment (Oxymoron79) [reply api/cloud/oc_cloud_manager.c:132]: I added the check in the commit https://github.com/iotivity/iotivity-lite/pull/192/commits/124049a76267ef0afe2130dab07bb4c1e49f4c3e

File: api/cloud/oc_cloud_manager.c [hunk 2]
```diff
@@ -123,7 +124,13 @@ cloud_start_process(oc_cloud_context_t *ctx)
   ctx->retry_count = 0;
   ctx->retry_refresh_token_count = 0;
 
-  if (ctx->store.status == OC_CLOUD_INITIALIZED) {
+  oc_sec_pstat_t *pstat = oc_sec_get_pstat(ctx->device);
```
Danielius1922, line 127: `#ifdef OC_SECURITY` around lines 127-130 is missing

PR #194, opened 2022-02-21
Minor updates to general_post

FrankZRS:


FrankZRS:
~format

security/oc_obt.c
```diff
@@ -3529,6 +3529,19 @@ oc_obt_general_post(oc_uuid_t *uuid, char *query, const char *url,
         } else {
           cbor_encode_text_string(&root_map, "", 0);
         }
+      } else if (strstr(payload_types[i], "bytes") != NULL) {
+        int byte_string_len = (strlen(payload_values[i])  + 1) / 2;
+        unsigned char payload_byte_string[byte_string_len];
```
WAvdBeek, line 3534: this line does not compile on windows..

python/iotivity.py
```diff
@@ -1157,7 +1158,7 @@ def offboard_all_owned(self):
 
             run_count = 0
             result = False
-            while run_count < 5 and not result: 
+            while run_count < 1 and not result:
```
Danielius1922, line 1161: No need for while, this will now always only run once

FrankZRS: Thanks for that!

WAvdBeek:
~format

WAvdBeek (approved):

WAvdBeek:
~format

WAvdBeek:
not sure why the commit in the format action is not working, merging it anyway, and then fix it later

PR #202, opened 2022-02-24
linux: flush system buffers after writing to storage

Danielius1922:


port/linux/storage.c
```diff
@@ -80,8 +81,10 @@ oc_storage_write(const char *store, uint8_t *buf, size_t size)
   if (!fp)
     return -EINVAL;
 
-  size = fwrite(buf, 1, size, fp);
+  size_t wsize = fwrite(buf, 1, size, fp);
```
Danielius1922, line 84: @jkralik do we want to handle all possible errors here? 

* size != wsize (when `fwrite` fails)
* `fflush`, `fsync`, `fileno`, `fclose` failures

I guess we should.

jkralik: yes we want

port/linux/storage.c
```diff
@@ -80,8 +81,10 @@ oc_storage_write(const char *store, uint8_t *buf, size_t size)
...
+  fflush(fp);
+  fsync(fileno(fp));
```
Oxymoron79, line 86: `fsync` is not available on Windows.

jkralik: But this is `port/linux` :)

jkralik: For windows we can use [_commit](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/commit?view=msvc-170)

jkralik: emulate [fsync](http://ab-initio.mit.edu/octave-Faddeeva/gnulib/lib/fsync.c)

Oxymoron79: > But this is port/linux :)
Your're right - I somehow missed that...

port/linux/storage.c
Danielius1922: yes, there's not much to do when `fclose`/`close` fail, I guess we can log it.

port/linux/storage.c
```diff
@@ -64,24 +67,53 @@ oc_storage_read(const char *store, uint8_t *buf, size_t size)
...
-  fclose(fp);
-  return size;
+  while (true) {
+    FILE *fp = fopen(store_path, "wb");
+    if (!fp) {
+      return -EINVAL;
+    }
+
+    long ret = write_and_flush(fp, buf, size);
+    if (fclose(fp) != 0) {
```
jkralik, line 110: this can lead to leak `fd` for `EAGAIN` and `EINTR`

Danielius1922: yeah, good catch, thanks.

Danielius1922: There seems to be not much we can do in case `fclose` or `close` fail. https://stackoverflow.com/questions/31527706/proper-error-handling-for-fclose-impossible-according-to-manpage

jkralik: Ok. So if `fclose` failes we need to break `while` cycle.

jkralik (approved):

WAvdBeek (approved):

ondrejtomcik:
@Danielius1922 could you please create github issue which would cover fix for Windows?

Danielius1922:
> @Danielius1922 could you please create github issue which would cover fix for Windows?

https://github.com/iotivity/iotivity-lite/issues/219

PR #205, opened 2022-02-28
Add payload encoder from text plain to cbor format

ibraprog:


CMakeLists.txt
```diff
@@ -158,11 +167,11 @@ file(GLOB PYTHON_SRC python/*.c)
 ######## Define link dependencies ########
 set(PRIVATE_LINK_LIBS "")
 
-include(CheckLibraryExists)
-check_library_exists(m ceil "" HAVE_LIB_M)
-if(HAVE_LIB_M)
+#include(CheckLibraryExists)
```
Danielius1922, line 170: Is there a problem with lines 170-174, or why are they commented out?

include/ausy_encoder.h
```diff
@@ -0,0 +1,11 @@
+#ifndef AUSY_ENCODER_H
```
Danielius1922, line 1: Maybe we should put a template for a public header somewhere, I usually copy it from some other file, but we have different version over the code base, so I am not sure which copyrights/trademarks are still relevant and should be included in new files. :confused:

api/ausy_encoder.c
```diff
@@ -0,0 +1,128 @@
+#include "ausy_encoder.h"
+#include "deps/json-parser/json.h"
+#include "oc_log.h"
+#include "deps/tinycbor/src/cborjson.h"
+
+static CborError ausy_encode_value_2_cbor(json_value* value, CborEncoder* currentEncoder);
```
Danielius1922, line 6: This declaration is not necessary, since the function has a definition with declaration on line 8

api/ausy_encoder.c
```diff
@@ -0,0 +1,128 @@
...
+{
+    CborError err = CborNoError;
+    int length, x;
+    if (value == NULL) {
+        return err;
+    }
+    length = value->u.object.length;
+    for (x = 0; x < length; x++) {
+        const char* key = value->u.object.values[x].name;
+        err |= cbor_encode_text_stringz(currentEncoder, key);
```
Danielius1922, line 18: I would break once the first error is encountered and print error (`OC_ERR`) with `key` so we know which attribute caused the problem

api/ausy_encoder.c
```diff
@@ -0,0 +1,128 @@
...
+static CborError ausy_encode_array_2_cbor(json_value* value, CborEncoder* currentEncoder)
+{
+    CborError err = CborNoError;
+    int length, x;
+    if (value == NULL) {
+        return err;
+    }
+    length = value->u.array.length;
+    for (x = 0; x < length; x++) {
+        err |= ausy_encode_value_2_cbor(value->u.array.values[x], currentEncoder);
```
Danielius1922, line 34: The same as https://github.com/iotivity/iotivity-lite/pull/205/files#r863579052

Danielius1922:
This seems to be abandoned.

PR #209, opened 2022-03-03
fixing csr buffer and enforce psk length during add

jkralik:
- possible override stack in CSR
- enforce PSK length according [specification](https://openconnectivity.org/specs/OCF_Security_Specification_v2.2.5.pdf) - 13.3.3.1 Symmetric key formatting

### DIFF
--- a/security/oc_cred.c
+++ b/security/oc_cred.c
@@ -53,6 +53,22 @@ static const char *allowed_roles[] = { "oic.role.owner" };
 static const int allowed_roles_num = sizeof(allowed_roles) / sizeof(char *);
 #endif
 
+// https://openconnectivity.org/specs/OCF_Security_Specification_v2.2.5.pdf
+// 13.3.3.1 Symmetric key formatting
+#define SYMMETRIC_KEY_128BIT_LEN 16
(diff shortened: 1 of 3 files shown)


security/oc_cred.c
```diff
@@ -552,13 +557,22 @@ oc_sec_add_new_cred(size_t device, bool roles_resource, oc_tls_peer_t *client,
       uint8_t key[64];
       memcpy(key, privatedata, privatedata_size);
       int key_size = oc_base64_decode(key, privatedata_size);
-      if (key_size < 0) {
+      if (key_size != SYMMETRIC_KEY_128BIT_LEN &&
```
Danielius1922, line 560: Maybe a helper function for the key size check and use it here and on lines 570-571.

Danielius1922 (approved):

security/oc_cred.c
jkralik: fixed

ondrejtomcik (approved):

WAvdBeek (approved):

WAvdBeek:
note: the csr size when from 4k to 0.5k so this might be reverted when the OCTT fails.

PR #211, opened 2022-03-04
coap: don't call memmove with empty source

Danielius1922:


messaging/coap/coap.c
```diff
@@ -1197,8 +1197,8 @@ coap_oscore_serialize_message(void *packet, uint8_t *buffer, bool inner,
     if (coap_pkt->payload_len > 0) {
       *option = 0xFF;
       ++option;
+      memmove(option, coap_pkt->payload, coap_pkt->payload_len);
```
Danielius1922, line 1200: GCC signature has nonnull annotations for the first and second parameter.
```
extern void *memmove (void *__dest, const void *__src, size_t __n)
     __THROW __nonnull ((1, 2));
```

So calling `memmove` with empty payload is undefined behavior.

jkralik (approved):

WAvdBeek (approved):

PR #212, opened 2022-03-04
Add functions to public API

Danielius1922:
Make oc_sec_get_creds, oc_sec_get_cred_by_credid and oc_sec_remove_cred_by_credid functions public.

### DIFF
--- a/include/oc_cred.h
+++ b/include/oc_cred.h
@@ -19,6 +19,7 @@
 #ifndef OC_CRED_COMMON_H
 #define OC_CRED_COMMON_H
 
+#include "oc_export.h"
 #include "oc_ri.h"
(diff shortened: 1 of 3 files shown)


Danielius1922:
@jkralik we also use the `post_cred` function but that doesn't seem like a good public function. I guess I'll have to create a new function with a better signature, or do you have a better idea?

jkralik:
> @jkralik we also use the `post_cred` function but that doesn't seem like a good public function. I guess I'll have to create a new function with a better signature, or do you have a better idea?

Im ok with the public funcs.

jkralik (approved):

WAvdBeek (approved):

PR #216, opened 2022-03-10
Add oc_store dump and load functions to public API

Danielius1922:


jkralik (approved):

include/oc_store.h
```diff
@@ -16,29 +16,111 @@
...
 extern "C" {
 #endif
 
-void oc_sec_load_pstat(size_t device);
-void oc_sec_load_doxm(size_t device);
-void oc_sec_load_cred(size_t device);
+/**
+ * @brief Load device acls from storage.
+ */
+OC_API
```
WAvdBeek, line 30: the doxygen documentation is not correct.
it is missing the description of the arguments
this also applies for all changes below

hint: use visual studio code with a plugin to create doxygen, e.g. that creates the body of the doxygen comment including all arguments

WAvdBeek (approved):

PR #217, opened 2022-03-11
Update gtest to v1.11.0 release

Danielius1922:


Danielius1922:
GTest no longer supports Makefile, so to use updated GTest we must first fully switch port/linux to CMake.

PR #223, opened 2022-03-17
Add public function to parse ACLs

Danielius1922:
Add oc_sec_apply_acl function to add ACLs to device by parsing them from a payload retrieved from some endpoint.

### DIFF
--- a/security/oc_acl.c
+++ b/security/oc_acl.c
@@ -898,7 +898,9 @@ oc_sec_acl_default(size_t device)
 }
 
 bool
-oc_sec_decode_acl(oc_rep_t *rep, bool from_storage, size_t device)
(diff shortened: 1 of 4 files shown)


security/oc_acl.c
```diff
@@ -1123,6 +1123,18 @@ oc_sec_acl_add_bootsrap_acl(size_t device)
 #endif
 }
 
+int
+oc_sec_apply_acl(oc_rep_t *rep, size_t device, bool dumpToStorage)
```
jkralik, line 1127: similar to creds -> we want to setup callback and user_data for each new cred

jkralik (approved):

WAvdBeek (approved):

PR #229, opened 2022-03-29
Update swupdate resource

Oxymoron79:
I would like to propose some changes to the implementation of the software update resource:

- Update the `nv` property in `oc_swupdate_notify_new_version_available`
- Clear the `nv` property in `oc_swupdate_notify_done`
- Perform follow-up action only on success in `oc_swupdate_notify_*` functions, i.e. check the result from the registered implementation before performing the follow-up action automatically.
- Remove the string length restriction for `purl`: IMO 63 characters is too low.
- Allow setting the `purl` property without the need to provide `swupdateaction` and `updatetime` in request: Setting the `purl` is not a timed action, more a configuration step.
- As the software update resource is a DCR, the URL should be prefixed with `oic/` as other DCR. And IMO `swu` is a better name for the resource. @SiMet: Would this change impact the certification tests?

### DIFF
--- a/api/oc_swupdate.c
+++ b/api/oc_swupdate.c
@@ -508,10 +508,6 @@ post_sw(oc_request_t *request, oc_interface_mask_t interfaces, void *user_data)
     if (oc_string_len(rep->name) == 4 &&
         memcmp(oc_string(rep->name), "purl", 4) == 0) {
       if (rep->type != OC_REP_STRING) {
-
-        error_state = true;
-      }
-      if (oc_string_len(rep->value.string) >= 63) {
         error_state = true;
       }
       purl = oc_string(rep->value.string);
--- a/api/oc_swupdate.c
+++ b/api/oc_swupdate.c
@@ -179,15 +179,13 @@ oc_swupdate_notify_new_version_available(size_t device, const char *version,
   oc_new_string(&s->nv, version, strlen(version));
   s->swupdatestate = OC_SWUPDATE_STATE_NSA;
   s->swupdateresult = result;
-  if(result != OC_SWUPDATE_RESULT_SUCCESS)
-  {
+  if(result != OC_SWUPDATE_RESULT_SUCCESS) {
     s->swupdateaction = OC_SWUPDATE_IDLE;
   }
 #ifdef OC_SERVER
   oc_notify_observers(oc_core_get_resource_by_index(OCF_SW_UPDATE, device));
 #endif /* OC_SERVER */
-  if(result == OC_SWUPDATE_RESULT_SUCCESS)
-  {
+  if(result == OC_SWUPDATE_RESULT_SUCCESS) {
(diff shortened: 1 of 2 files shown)


api/oc_swupdate.c
```diff
@@ -455,7 +467,7 @@ oc_swupdate_decode(oc_rep_t *rep, size_t device)
 }
 
 /**
- * post method for "/sw" resource.
```
WAvdBeek, line 458: why the oic part?
that is not needed, it makes it just a longer url

Oxymoron79: The software update resource is implemented as a device configuration resource (DCR). The URLs of most other DCRs are prefixed with `oic`. So I thought it's a kind of grouping for DCRs that iotivity-lite does and I propose to adjust the software update resource to follow that scheme.

It is not necessary, I just think it makes the DCR URLs more consistent.

jkralik (approved):

api/oc_swupdate.c
jkralik: I want to mentioned that we have resource `/oc/con`  and it has  comment that it cannot be changed: https://github.com/iotivity/iotivity-lite/blob/master/api/oc_core_res.c#L54

WAvdBeek: why change it?
this does not make the stack more consistent... 
the /oic/res etc are fixed by the spec the /sw is not.. so do not use /oic/ as a prefix, it would hint that that is a url specified by the spec

Oxymoron79: > the /oic/res etc are fixed by the spec the /sw is not.. so do not use /oic/ as a prefix, it would hint that that is a url specified by the spec

Thanks for the explanation @WAvdBeek. I agree that `/oic/` is the wrong prefix.

But I would still prefer to change it to a `/oc` prefix: In my opinion, all resources that the iotivity-lite stack implements with a hard-coded URL should have the common URL prefix `/oc` (like the `/oc/con` or `/oc/wk/introspection` resources).  That's similar to the best practice in REST API design where the resources are grouped with such prefixes (that's probably where my preference comes from).

And what about the resource name `sw`? To me that acronym just stands for "software". `swu` or even `swupd` (as the spec names as a example) would be more expressive, in my opinion.

WAvdBeek: oc/swu is fine by me

Oxymoron79: Changed the URL to `/oc/swu`.

WAvdBeek (approved):

SiMet:
Hi all,

I see that @Oxymoron79 asked me about impact the certification tests.
The answer is: Yes, it is.

We get 2 failing tests (1.2.3 and 1.7.8.3). Please see attachment 
[2022-04-05-00-22-27_Client Certification_retest.zip](https://github.com/iotivity/iotivity-lite/files/8415786/2022-04-05-00-22-27_Client.Certification_retest.zip)

PR #235, opened 2022-04-05
Update iotivity-lite port for Zephyr 3.0.0

Oxymoron79:
Update the port of iotivity-lite for the latest Zephyr RTOS (3.0.0).

- [x] Build with CMake
- [x] Update TinyCBOR to version 0.6 to be able to disable floating point support with Zephyr's minimal libc
- [x] Implement iotivity-lite storage using Zephyr's Filesystem module.
- [x] Implement IPv4/6 multicast handler
- [x] Port UDP handler using the Posix socket library (similar to the other iotivity-lite ports)
- [x] Implement network monitor
- [ ] Add atomic implementation for `#if defined(ZEPHYR)` to util/oc_atomic.h
- [ ] Implement client funtionality by implementing `oc_send_discovery_request`
- [ ] IPv4 DHCP client works only with the Zephyr network auto configuration (initializes the network interfaces before calling main()). Running the DHCP client in the application seems to lead to a deadlock. Further investigation is needed.
- [ ] With security enabled, the ownership transfer did not work. Further investigation is needed.


Oxymoron79:
~format

WAvdBeek:
Update TinyCBOR to version 0.6 to be able to disable floating point support with Zephyr's minimal libc

can we do this in a seperate pull request, e.g. to check if anything breaks?

Oxymoron79:
> can we do this in a seperate pull request, e.g. to check if anything breaks?

@WAvdBeek: Yes, sure. See #238

jkralik:
@Oxymoron79 Please, could you add example server + README which uses zephyr ?

Oxymoron79:
> @Oxymoron79 Please, could you add example server + README which uses zephyr ?

@jkralik I have now added a README for the Zephyr port and how I used it. Sorry that it took so long - I didn't find the time to write it...

PR #238, opened 2022-04-07
Update tinycbor submodule to v0.6.0

Oxymoron79:
TinyCBOR 0.6.0 adds compilation flags that may be needed for a future
update of the Zephyr port (#235).


SiMet:
CTT testing passed. There is no failure found.

jkralik:
Could you update Makefile's in ports? If someone use them then it cause the issue, because the cbor*_float.c files aren't set.

Oxymoron79:
@jkralik I updated the Makefile for the other ports.

jkralik (approved):

WAvdBeek (approved):

PR #255, opened 2022-05-03
Py cmd arg

FrankZRS:
Added command line argument parsing to iotivity.py; 

Added general DELETE requesting sending functions; 

Created GUI for the proxy mediator.


FrankZRS:
~format

WAvdBeek (approved):

PR #264, opened 2022-05-16
Add oc_cred_get_validity_period function

Danielius1922:
Add public function to parse valid_from and valid_to times from the certificate associated with the credential.


security/oc_cred.c
```diff
@@ -1066,6 +1071,78 @@ oc_cred_parse_credusage(oc_string_t *credusage_string)
...
+bool
+oc_cred_get_validity_period(const oc_sec_cred_t *cred, uint64_t *valid_from,
+                            uint64_t *valid_to)
+{
+  oc_assert(cred != NULL);
+  oc_assert(valid_from != NULL);
+  oc_assert(valid_to != NULL);
+
+  mbedtls_x509_crt crt;
+  if (!oc_cred_parse_certificate(cred, &crt)) {
```
jkralik, line 1135: is it possible to avoid parsing(allocate) memory? Maybe this crt is already parsed in some list oc_tls.c.

jkralik (approved):

PR #265, opened 2022-05-19
update mbedtls to 3.1.0

jkralik:


Danielius1922:
Android Makefile needs to be updated. (We really need to get rid of Makefiles, so we don't have to maintain two build systems.)

jkralik:
@SiMet Pls could you look at why CTT fails? In this MR I updated `deps/mbedtls` to 3.1.0. Thx

SiMet:
> @SiMet Pls could you look at why CTT fails? In this MR I updated `deps/mbedtls` to 3.1.0. Thx

Hi,

There was issue with update submodules. I've added --force option to 'git submodule update --init --recursive'. So I hope now should be fine. I am not able to test it now because other test run is in progress.

Danielius1922 (approved):

PR #269, opened 2022-05-23
add Push Notification support (CR2483)

joochlee:
Push Notification support for iotivity-lite (CR2483)


joochlee:

include/oc_api.h
```diff
@@ -1134,6 +1134,12 @@ void oc_resource_make_public(oc_resource_t *resource);
  */
 void oc_resource_set_discoverable(oc_resource_t *resource, bool state);
 
+
+#if defined(OC_PUSH) && defined(OC_SERVER) && defined(OC_CLIENT) && defined(OC_DYNAMIC_ALLOCATION) && defined(OC_COLLECTIONS_IF_CREATE)
+void oc_resource_set_pushable(oc_resource_t *resource, bool state);
```
WAvdBeek, line 1139: please add doxygen documentaton on the function

see also other header files

include/oc_ri.h
```diff
@@ -50,6 +50,9 @@ typedef enum {
 typedef enum {
   OC_DISCOVERABLE = (1 << 0), ///< discoverable
   OC_OBSERVABLE = (1 << 1),   ///< observable
+#if defined(OC_PUSH) && defined(OC_SERVER) && defined(OC_CLIENT) && defined(OC_DYNAMIC_ALLOCATION) && defined(OC_COLLECTIONS_IF_CREATE)
+  OC_PUSHABLE = (1 << 2),
```
WAvdBeek, line 54: add doxygen

port/linux/Makefile
```diff
@@ -194,6 +194,16 @@ ifeq ($(SWUPDATE),1)
 	export SWUPDATE
 endif
 
+
```
WAvdBeek, line 197: no other makefiles e.g. windows?

.project
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
```
WAvdBeek, line 2: do we need this file?

.gitignore
```diff
@@ -111,3 +111,7 @@ service/resource-directory/client/unittest/obj
 build
 
 python/plgd_headers.config
+
+# eclipse file
+.cproject
```
WAvdBeek, line 116: hmm, git ignore does not want .project but the file is just below..

.cproject
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
WAvdBeek, line 1: i guess this file also needs to be removed, since .gitignore filters this one out.

api/oc_push.c
```diff
@@ -0,0 +1,2580 @@
...
+
+/*
+ * if this callback function is provided by user, it will called whenever new push is arrived...
+ */
+void (*oc_push_arrived)(oc_pushd_rsc_rep_t *) = NULL;
+
+
+
+/*
+ * FIXME4ME use `oc_ri_get_interface_mask()` instead of this...
```
WAvdBeek, line 161: not sure if this still needs to be fixed

api/oc_push.c
```diff
@@ -0,0 +1,2580 @@
...
+	inet_ntop(AF_INET, ns_instance->pushtarget_ep.addr.ipv4.address, ipv4addrstr, 50);
+
+	if (!strcmp(ipv6addrstr, "::") && !strcmp(ipv4addrstr, "0.0.0.0"))
+		return true;
+
+	return false;
+}
+
+
+/**
```
WAvdBeek, line 218: the doxygen annotation should be in the header files.

WAvdBeek (changes requested):

include/oc_push.h
```diff
@@ -0,0 +1,203 @@
...
+
+#define oc_init_string(str) \
+{ \
+	(str).size = 0; \
+	(str).ptr = NULL; \
+	(str).next = NULL; \
+}
+
+
+void oc_push_list_init();
```
WAvdBeek, line 177: please add doxygen to all functions

include/oc_push.h
```diff
@@ -0,0 +1,203 @@
...
+} oc_recvs_t;
+
+
+
+/*
+ * object used to store Resource pushed to "oic.r.pshreceiver:receivers[i].receiveruri"
+ */
+typedef struct oc_pushd_rsc_rep
+{
+	struct oc_pushd_rsc_rep *next;
```
WAvdBeek, line 137: add doxygen to the fields. 
for example 

/**
 * @brief device information
 *
 */
typedef struct oc_device_info_t
{
  oc_uuid_t di;                          ///< device indentifier
  oc_uuid_t piid;                        ///< Permanent Immutable ID
  oc_string_t name;                      ///< name of the device
  oc_string_t icv;                       ///< specification version
  oc_string_t dmv;                       ///< data model version
  oc_core_add_device_cb_t add_device_cb; ///< callback when device is changed
  void *data;                            ///< user data
} oc_device_info_t;

WAvdBeek (changes requested):

.github/workflows/cmake-linux.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           - args: "-DOC_IPV4_ENABLED=ON"
           # insecure with IPv4
           - args: "-DOC_SECURITY_ENABLED=OFF -DOC_IPV4_ENABLED=ON"
+          # enable Push Notification
+          - args: "-DOC_COLLECTIONS_IF_CREATE_ENABLED=ON -DOC_PUSH_ENABLED=ON -DOC_DYNAMIC_ALLOCATION_ENABLED=ON"
```
Danielius1922, line 26: `OC_DYNAMIC_ALLOCATION_ENABLED` is enabled by default, so you can remove it to simplify the expression.

CMakeLists.txt
```diff
@@ -148,6 +150,13 @@ if(OC_DNS_LOOKUP_IPV6_ENABLED)
     list(APPEND PRIVATE_COMPILE_DEFINITIONS "OC_DNS_LOOKUP_IPV6")
 endif()
 
+if(OC_PUSH_ENABLED)
```
Danielius1922, line 153: From the `#ifdef` I deduce that `OC_PUSH` required to `OC_DYNAMIC_ALLOCATION` and `OC_COLLECTIONS_IF_CREATE`. In that case I would copy how `OC_CLOUD_ENABLED` is handled (eg. it forces the required options to be enabled):
```
if(OC_CLOUD_ENABLED)
    list(APPEND PUBLIC_COMPILE_DEFINITIONS "OC_CLOUD")
    set(OC_TCP_ENABLED ON)
    set(OC_IPV4_ENABLED ON)
endif()
```

So I would change this to:
```
if(OC_PUSH_ENABLED)
    list(APPEND PRIVATE_COMPILE_DEFINITIONS "OC_PUSH")
    if(OC_PUSHDEBUG_ENABLED)
        list(APPEND PRIVATE_COMPILE_DEFINITIONS "OC_PUSHDEBUG")
    endif()
    set(OC_DYNAMIC_ALLOCATION_ENABLED ON)
    set(OC_COLLECTIONS_IF_CREATE_ENABLED ON)
endif()
```
And these lines have to be moved before the forced variables are first used - that means before line 87 where `OC_DYNAMIC_ALLOCATION_ENABLED` is checked.

include/oc_push.h
```diff
@@ -0,0 +1,232 @@
...
+#if defined(OC_PUSHDEBUG) || defined(OC_DEBUG)
+#ifndef OC_DEBUG
+#define oc_print(...) printf(__VA_ARGS__)
+#define oc_log(level, ...)                                                     \
+		do {                                                                         \
+			oc_print("%s: %s <%s:%d>: ", level, __FILE__, __func__, __LINE__);            \
+			oc_print(__VA_ARGS__);                                                        \
+			oc_print("\n");                                                               \
+		} while (0)
+#define p_dbg(...) oc_log("DEBUG", __VA_ARGS__)
```
Danielius1922, line 47: Lets keep the convention that macros are always all upper case. (I see that some log macros have small letter suffixes - such as `PRINTipaddr` - I'll fix that eventually. To keep other conventions it should be something like

p_dbg -> OC_PUSH_DBG
p_wrn -> OC_PUSH_WRN
p_err -> OC_PUSH_ERR

include/oc_push.h
```diff
@@ -0,0 +1,232 @@
...
+#include "oc_endpoint.h"
+#include "port/oc_log.h"
+#include "util/oc_memb.h"
+#include "util/oc_process.h"
+
+
+#if defined(OC_PUSHDEBUG) || defined(OC_DEBUG)
+#ifndef OC_DEBUG
+#define oc_print(...) printf(__VA_ARGS__)
+#define oc_log(level, ...)                                                     \
```
Danielius1922, line 41: This is the same macro as in `port/oc_log.h`, right? In that case would reuse the `OC_LOG` define from there and modify the conditions there so it is defined for `OC_PUSHDEBUG`:

Changes to `port/oc_log.h`
```
#if defined(OC_DEBUG) || defined(OC_PUSHDEBUG)
#ifdef __ANDROID__
#define OC_LOG(level, ...)                                                     \
  android_log(level, __FILE__, __func__, __LINE__, __VA_ARGS__)
...
#endif /* __ANDROID__ */
#endif /* OC_DEBUG || OC_PUSHDEBUG */

#ifdef OC_DEBUG
#define OC_DBG(...) OC_LOG("D", __VA_ARGS__)
#define OC_WRN(...) OC_LOG("W", __VA_ARGS__)
#define OC_ERR(...) OC_LOG("E", __VA_ARGS__)
```

So now the `OC_LOG` would be defined whenever at least one of `OC_DEBUG` or `OC_PUSHDEBUG` is defined. `OC_LOG` can be used in `oc_push.h` and `OC_DBG`/`OC_WRN`/`OC_ERR` are not defined if `OC_DEBUG` is not defined.

include/oc_push.h
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * @brief initiate Push Receiver Resource
+ */
+void oc_create_pushreceiver_resource(size_t device_index);
+
+/**
+ * @brief try to find `receiver` object which has `uri` as its `uri` Property
+ */
+oc_recv_t * _find_recv_obj_by_uri(oc_recvs_t *recvs_instance, const char *uri, int uri_len);
```
Danielius1922, line 205: This doesn't seem like a public function, so we shouldn't expose the declaration in a public header. It seems to be used in just `oc_push.c` so just add the `static` keyword to the definition in the `oc_push.c` file.

include/oc_push.h
```diff
@@ -0,0 +1,232 @@
...
+ *
+ * @param[in] uri          path of pushable Resource whose contents is just updated
+ * @param[in] device_index index of Device that updated pushable Resource belongs to
+ */
+void oc_resource_state_changed(const char *uri, size_t device_index);
+
+/**
+ * @brief try to find `receiver` object which has `uri_string` as its `uri` Property
+ */
+#define _find_recv_obj_by_uri2(recvs_instance, uri_string) \
```
Danielius1922, line 224: Only used in `oc_push.c`, right? So please move this to that file.

include/oc_push.h
```diff
@@ -0,0 +1,232 @@
...
+{
+	struct oc_pushd_rsc_rep *next;
+	oc_resource_t *resource;        ///< used to point any pushed Resource managed by iotivity-lite
+	oc_rep_t *rep;                  ///< payload of pushed Resource
+} oc_pushd_rsc_rep_t;
+
+
+OC_PROCESS_NAME(oc_push_process);
+
+extern char *pp_state_strs[];
```
Danielius1922, line 148: In public headers we should only expose what is necessary. I think this and `cli_status_strs`, `cli_statusstr`, `pp_statestr`, `pp_update_state`, `oc_init_string` are all only used in `oc_push.c`, so all these definitions should be moved there. Or into a `oc_push_internal.h` header if they are reused in some other internal implementation file.

include/oc_api.h
joochlee: done

CMakeLists.txt
joochlee: Thanks, fixed !

api/oc_ri.c
```diff
@@ -373,6 +378,14 @@ start_processes(void)
 #ifdef OC_TCP
   oc_process_start(&oc_session_events, NULL);
 #endif /* OC_TCP */
+
+  /*
+   * TODO4ME<done> start oc_push_process
```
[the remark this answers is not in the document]
joochlee: fixed

include/oc_push.h
joochlee: I agree, fixed !

include/oc_push.h
joochlee: fixed

include/oc_push.h
joochlee: fixed

include/oc_push.h
joochlee: fixed

include/oc_ri.h
joochlee: done

.project
joochlee: fixed

.cproject
joochlee: fixed

api/oc_push.c
joochlee: fixed

api/oc_push.c
joochlee: done

include/oc_push.h
joochlee: done

api/oc_push.c
```diff
@@ -0,0 +1,2658 @@
...
+#define oc_init_string(str) \
+{ \
+	(str).size = 0; \
+	(str).ptr = NULL; \
+	(str).next = NULL; \
+}
+
+
+
+static oc_interface_mask_t _get_ifmask_from_ifstr(char *ifstr)
```
Danielius1922, line 246: Already existing `oc_ri_get_interface_mask` can be used instead of this new function

.github/workflows/cmake-linux.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           - args: "-DOC_IPV4_ENABLED=ON"
           # insecure with IPv4
           - args: "-DOC_SECURITY_ENABLED=OFF -DOC_IPV4_ENABLED=ON"
+          # enable Push Notification
+          - args: "-DOC_COLLECTIONS_IF_CREATE_ENABLED=ON -DOC_PUSH_ENABLED=ON"
```
Danielius1922, line 26: Do you see the Checks tab, you already have a run for this setup - https://github.com/iotivity/iotivity-lite/runs/6665915581?check_suite_focus=true . :) It failed but that's not your fault, we have a problematic default setup in CMake (I'm trying to fix the default setup in https://github.com/iotivity/iotivity-lite/pull/272), so for now add `-DOC_TCP_ENABLED=OFF` and it should succeed.

api/oc_push.c
```diff
@@ -0,0 +1,2658 @@
...
+
+	return iface;
+}
+
+
+static bool _is_null_pushtarget(oc_ns_t *ns_instance)
+{
+	char ipv6addrstr[50], ipv4addrstr[50];
+
+	inet_ntop(AF_INET6, ns_instance->pushtarget_ep.addr.ipv6.address, ipv6addrstr, 50);
```
Danielius1922, line 291: I'm not sure `inet_ntop` is guaranteed not to write nonsense into the output buffer in case of an error, it would be safer to check the return value. Maybe something like this?

```C
if (inet_ntop(AF_INET6, ns_instance->pushtarget_ep.addr.ipv6.address, ipv6addrstr, sizeof(ipv6addrstr)) != NULL && strcmp(ipv6addrstr, "::") != 0) {
  return false;
}

if (inet_ntop(AF_INET, ns_instance->pushtarget_ep.addr.ipv4.address, ipv4addrstr, sizeof(ipv4addrstr)) != NULL && strcmp(ipv4addrstr, "0.0.0.0") != 0) {
  return false;
}

return true;
```

joochlee:
~format

include/oc_push.h
```diff
@@ -0,0 +1,111 @@
...
+#include "oc_helpers.h"
+#include "oc_rep.h"
+#include "oc_ri.h"
+#include "oc_endpoint.h"
+#include "port/oc_log.h"
+#include "util/oc_memb.h"
+#include "util/oc_process.h"
+
+
+#if defined(OC_PUSHDEBUG) || defined(OC_DEBUG)
```
Danielius1922, line 38: Can't lines 38-46 be moved into `oc_push.c` file? This seems like an internal logging of the push mechanism, doesn't need to be public.

include/oc_push.h
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * @brief return response code string for `i`
+ */
+#define cli_statusstr(i) (cli_status_strs[(i)])
+
+
+/**
+ * @brief print payload of Resource in user friendly format
+ */
+void oc_print_pushd_rsc(oc_rep_t *payload);
```
Danielius1922, line 95: I think payload can be made `const`, it only prints right?

include/oc_push.h
```diff
@@ -0,0 +1,111 @@
...
+	struct oc_pushd_rsc_rep *next;
+	oc_resource_t *resource;        ///< used to point any pushed Resource managed by iotivity-lite
+	oc_rep_t *rep;                  ///< payload of pushed Resource
+} oc_pushd_rsc_rep_t;
+
+
+/**
+ * @brief string representing response code
+ */
+extern const char *cli_status_strs[];
```
Danielius1922, line 79: We should avoid global variables if possible:

1) can't cli_status_strs and cli_statusstr be moved to oc_push.c? I don't see them being used outside of that file. If they need to be in a header then they shouldn't be in a public header (eg. a header in the include directory, these headers are copied to /usr/local/lib/iotivity-lite when somebody install the library and thus should feature only public classes and functions).

2) oc_push_arrived -> make the declaration of the variable static in the oc_push.c and create a setter function:
```C
typedef void (*oc_on_push_arrived_t)(oc_pushd_rsc_rep_t *);

void oc_set_on_push_arrived(oc_on_push_arrived_t*);
```

(or something like that), which sets the variable.

include/oc_push.h
```diff
@@ -0,0 +1,111 @@
...
+void oc_print_pushd_rsc(oc_rep_t *payload);
+
+/**
+ * @brief application should call this function whenever the contents of pushable Resource is updated,
+ *        or Push Notification will not work.
+ *
+ * @param[in] uri          path of pushable Resource whose contents is just updated
+ * @param[in] device_index index of Device that updated pushable Resource belongs to
+ */
+void oc_resource_state_changed(const char *uri, size_t device_index);
```
Danielius1922, line 104: Please add `OC_API` annotation to all functions in this public (see for example: https://github.com/iotivity/iotivity-lite/blob/master/include/oc_cloud.h , it's needed for support of shared libraries/DLLs on Windows; all new public functions should have it and I'm working on adding it to old code and then we can enable compilation of DLLs)

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+/*
+ * @brief update Push Proxy state from state to new_state
+ *
+ * @param state		oc_string_t
+ * @param new_state	char *
+ */
+#define pp_update_state(state, new_state) \
+{ \
+	oc_free_string(&(state)); \
```
Danielius1922, line 230: To replace oc_string_t with a new value is a common code (for example I see it in many places in `set_ns_properties`). Let's add it to oc_helpers.h/.c:

```C
oc_set_string(oc_string_t *ocstring, const char *str, size_t str_len) {
  oc_free_string(ocstring);
  oc_new_string(ocstring, str, str_len);
}
```

Then you can either use `oc_set_string` directly or 
```C
#define pp_update_state(state, new_state) oc_set_string(&(state), (new_state), strlen((new_state)))
```

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+			/* free each field of ns_instance */
+			oc_free_string(&ns_instance->phref);
+			oc_free_string_array(&ns_instance->prt);
+			oc_free_string_array(&ns_instance->pif);
+
+			ep = ns_instance->pushtarget_ep.next;
+			while (ep)
+			{
+				oc_free_endpoint(ep);
```
Danielius1922, line 730: `oc_free_endpoint` deallocates all memory, including `next`, so you have to use a helper variable to avoid use of memory after free:

```C
  oc_endpoint_t *next;
  while (ep != NULL) {
    next = ep->next;
    oc_free_endpoint(ep);
    ep = next;
  }
```

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+/*
+ * @brief				check if "rt" of pushed resource is part of "rts" (all value of "rt" should be part of "rts")
+ *
+ * @param recv_obj
+ * @param rep
+ * @return	not 0: found
+ * 			0: not found
+ */
+static char _check_pushd_rsc_rt(oc_recv_t *recv_obj, oc_rep_t *rep)
```
Danielius1922, line 1066: (CheckPoint 1: checked this file upto this point, publishing my review so changes can be done, will read the rest tomorrow.)

Danielius1922 (changes requested):

include/oc_push.h
joochlee: done

include/oc_push.h
joochlee: right, fixed

include/oc_push.h
joochlee: 1. I'm using that macro in other apps (e.g. push configurator)
2. fixed

include/oc_push.h
joochlee: fixed

api/oc_push.c
joochlee: done

api/oc_push.c
joochlee: oops.. fixed...

api/oc_push.c
joochlee: oops.. fixed!

include/oc_helpers.h
```diff
@@ -94,6 +94,12 @@ typedef struct oc_mmem oc_handle_t, oc_string_t, oc_array_t, oc_string_array_t,
  */
 #define oc_free_string(ocstring) _oc_free_string(ocstring)
 
+/**
+ * @brief reset ocstring
+ *
+ */
+#define oc_set_string(ocstring, str, str_len) _oc_set_string(ocstring, str, str_len)
```
Danielius1922, line 101: I think this is not needed, just name the function `oc_set_string` (instead of `_oc_set_string`).

If you look at `oc_free_string` for example, the macro is there because of `OC_MEMORY_TRACE`. The call to `_oc_free_string` is different based on whether that macro is defined or not. But that is not the case for us here, so we don't need to macro override for `oc_set_string` and can define it as function directly.

joochlee: oh..  understood

include/oc_push.h
Danielius1922: 1. Would a public function work for you? Enum `oc_status_t` is defined in `oc_ri.h`, when I look into that header there is already a public function called `oc_status_code` which converts the status to some other code. Lets add a new public function `oc_status_to_str(oc_status_t)` after `oc_status_code` and you can use that in your app.

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+	/* create Push Receiver Resource */
+	oc_resource_t *pushd_rsc = oc_new_resource("Pushed Resource", oc_string(recv_obj->receiveruri), 1, resource->device);
+
+	if (pushd_rsc)
+	{
+		/*
+		 * XXX, if a resource binds empty resource type (""), when a client retrieve this it may receive weird value...
+		 */
+		oc_resource_bind_resource_type(pushd_rsc, " ");
```
Danielius1922: Why the empty space " ", isn't there some designated resource type in the specification?

@jkralik

include/oc_push.h
joochlee: done

api/oc_push.c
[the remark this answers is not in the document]
joochlee: fixed

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+
+
+/*
+ * @brief				oc_rep_set_pool() should be called before calling this func
+ *
+ * @param new_rep
+ * @param org_rep
+ */
+static void * _create_pushd_rsc_rep(oc_rep_t **new_rep, oc_rep_t *org_rep)
```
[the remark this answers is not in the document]
joochlee: done

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+	case OC_REP_BYTE_STRING:
+	case OC_REP_STRING:
+		oc_new_string(&((*new_rep)->value.string), oc_string(org_rep->value.string), oc_string_len(org_rep->value.string));
+		break;
+	case OC_REP_OBJECT:
+		(*new_rep)->value.object = _create_pushd_rsc_rep(&((*new_rep)->value.object), org_rep->value.object);
+		break;
+	case OC_REP_OBJECT_ARRAY:
+		/*
+		 * TODO4ME, test this!
```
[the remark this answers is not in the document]
joochlee: it has already been tested. I forgot to remove comment. 
fixed

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+#define _find_recv_obj_by_uri2(recvs_instance, uri_string) \
+	(_find_recv_obj_by_uri((recvs_instance), oc_string(uri_string), oc_string_len(uri_string)))
+
+
+
+static oc_rep_t * _rep_list_remove(oc_rep_t **rep_list, oc_rep_t **item)
+{
+	oc_rep_t **l, *removed_item;
+
+	for (l = (oc_rep_t**) rep_list; *l != NULL; l = &(*l)->next)
```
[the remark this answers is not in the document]
joochlee: oops.. fixed

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+	{
+		OC_PUSH_ERR("pushed resource type(s) is not in \"rts\" of push recerver object");
+		result = OC_STATUS_FORBIDDEN;
+	}
+	else
+	{
+		while (rep)
+		{
+			/*
+			 * <2022/4/20> skip "rt" (array), "if" (array), "n" (optional), "id" (optional) common property in the payload ("oic.r.pushpayload")
```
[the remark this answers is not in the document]
joochlee: nop, fixed

include/oc_ri.h
```diff
@@ -342,6 +360,11 @@ typedef struct oc_event_callback_s
   void *data;                       ///< data for the callback
 } oc_event_callback_t;
 
+/**
+ * @brief CoAP status code string
+ */
+extern const char *cli_status_strs[];
```
Danielius1922, line 366: You shouldn't need this since `oc_status_to_str` is available, right?

include/oc_ri.h
```diff
@@ -397,6 +420,16 @@ void oc_ri_remove_timed_event_callback(void *cb_data,
...
 int oc_status_code(oc_status_t key);
 
+
+/**
+ * @brief convert the status code to string
+ *
+ * @param[in] key key the application level key of the code
+ * @return char* CoAP status code string
+ */
+const char *oc_status_to_str(oc_status_t key);
```
Danielius1922, line 430: Add a `OC_API` annotation since it is public:

```
OC_API
const char *oc_status_to_str(oc_status_t key);
```

api/oc_push.c
```diff
@@ -0,0 +1,2555 @@
...
+ * @brief initialize oc_string_t object
+ */
+#define oc_init_string(str)                                                    \
+  {                                                                            \
+    (str).size = 0;                                                            \
+    (str).ptr = NULL;                                                          \
+    (str).next = NULL;                                                         \
+  }
+
+OC_API
```
Danielius1922, line 214: Does this work? I always thought the annotation needed to be in the header, but this way would be nicer actually, less clutter in the header files. I'm going to make a note and see if it works.

api/oc_push.c
```diff
@@ -0,0 +1,2555 @@
...
+  return recvs_instance;
+}
+
+/*
+ * @brief				oc_rep_set_pool() should be called before calling this func
+ *
+ * @param new_rep
+ * @param org_rep
+ */
+static void *
```
Danielius1922, line 1134: `static oc_rep_t*`, doesn't seem that you need the indeterminate void-ptr type

api/oc_push.c
```diff
@@ -0,0 +1,2555 @@
...
+ * @param user_data
+ */
+static void
+post_pushrecv(oc_request_t *request, oc_interface_mask_t iface_mask,
+              void *user_data)
+{
+  (void)iface_mask;
+  (void)user_data;
+
+  char *uri_param;
```
Danielius1922, line 1950: `char *uri_param = NULL;`, because when `request->query` is `NULL` this value will not get initialized by oc_ri_get_query_value and it will have random, undefined value.

include/oc_ri.h
joochlee: oops, right. 
fixed.

api/oc_push.c
joochlee: I think so. 
I found many example using visibility in function definition

include/oc_ri.h
joochlee: Fixed, 
do you think that `OC_API` should also be added to the new function such as `oc_set_string()` in `oc_helpers.*` ?

api/oc_push.c
joochlee: agree, fixed

api/oc_push.c
joochlee: agree, fixed

api/oc_push.c
Danielius1922: Indeed, seems to work on Unix (I've created a `.so` library and then invoked `nm -D` on it and the symbol was visible). Good to know.

However, I think this will be a problem on Windows with MSVC. If I understand it correctly, the exported function must be annocated with `dllExport` in the library itself and then with `dllImport` when the library is used in a binary. And this can only be done in the header.

joochlee: ok, I'll leave it in header only

include/oc_ri.h
Danielius1922: Yeah, it should have it, so played add it. 

I haven't got  around yet to updating old code, but we should add it when we create new functions, so we don't increase the amount of work necessary to make .DLLs on Windows work.

include/oc_push.h
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @brief print payload of Resource in user friendly format
+ */
+OC_API
+void oc_print_pushd_rsc(const oc_rep_t *payload);
+
+/**
+ * @brief set callback function called whenever new push arrives
+ *
+ * @param[in] function name
```
Danielius1922, line 75: I think doxygen expects the name of the parameter after the `param` statement

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+ * @brief						remove receiver object array in `recv_obj_list`,
+ * 								and app resource pointed by `receiveruri` of each receivre object in the array
+ *
+ * @param recv_obj_list		receiver object array
+ * @param device_index		index of device where the Push Resource resides
+ */
+static void _purge_recv_obj_list(oc_recvs_t *recvs_instance)
+{
+	/*
+	 * TODO4ME rewrite this func (refer to oc_free_rep())
```
Danielius1922: Is this (and other) TODO4ME still relevant? :-)

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+
+		default:
+			OC_PUSH_ERR("something wrong, unexpected Property type: %d", rep->type);
+			break;
+		}
+		rep = rep->next;
+	}
+
+	if (mandatory_property_check != 0x3)
+		return result;
```
Danielius1922: I think `recv_obj` leak here, it should be deallocated before returning

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+	else
+		result = true;
+
+	oc_list_add(recvs_instance->receivers, recv_obj);
+
+	/* create app resource corresponding to receiver object */
+	OC_PUSH_DBG("new app resource for new receiver obj (\"%s\") is created...", oc_string(recv_obj->receiveruri));
+	if (_create_pushd_rsc(recv_obj, recvs_instance->resource))
+		result = true;
```
Danielius1922: Isn't this inverted? You either return on line 1900 or go to the else branch, which sets `result=true`. So the value is already set to `true`. Maybe you should set the result to `false` in case it fails and log error?

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+			for (int i=0; i<obj_arr_len; i++, rep_obj=rep_obj->next)
+			{
+				if (!_create_recv_obj(recvs_instance, rep_obj->value.object))
+					result = false;
+			} /* for */
+			break;
+
+		default:
+			OC_PUSH_ERR("something wrong, unexpected Property type: %d", rep->type);
+			break;
```
Danielius1922: I think this should return false instead of just breaking. For example if the first `rep->type` is not `OC_REP_OBJECT_ARRAY` it will break and return true.

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+ * @param rep				payload representation of new receiver object array
+ */
+static bool _replace_recv_obj_array(oc_recvs_t *recvs_instance, oc_rep_t *rep)
+{
+	int obj_arr_len;
+	oc_rep_t *rep_obj;
+	bool result = true;
+
+	/* remove existing receivers object array */
+	_purge_recv_obj_list(recvs_instance);
```
Danielius1922: Wouldn't it be better to keep the recv_obj_list intact in case some invocation `_create_recv_obj` fails.

Usually the stuff is done in the following fashion:
1) prepare new data to some temporarily list (in this case this would be the `while` cycle, but it would write the data to a new oc_list)
2) if an error occurs deallocate stuff in the temp list, write error and return
3) if everything is in order -> deallocate the original list and replace it with the new list

Though this would potentially have higher memory requirements, because for a moment you'd hold both the old and new lists in memory. This shouldn't be a problem if the lists aren't big.

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+		 */
+		oc_recvs_t *recvs_instance = (oc_recvs_t *)oc_memb_alloc(&recvs_instance_memb);
+		if (recvs_instance) {
+			recvs_instance->resource = push_recv;
+			OC_LIST_STRUCT_INIT(recvs_instance, receivers);
+			oc_list_add(recvs_list, recvs_instance);
+		}
+		else
+		{
+			OC_PUSH_ERR("oc_memb_alloc() error!");
```
Danielius1922: Shouldn't the whole resource be deallocated in case this happens?

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+	}
+
+}
+
+
+
+/*
+ * alternative implementation (not using proto-thread)
+ */
+static void push_update(oc_ns_t *ns_instance)
```
Danielius1922: (Checkpoint 3)

include/oc_ri.h
joochlee: done

api/oc_push.c
joochlee: done

include/oc_push.h
joochlee: fixed

api/oc_push.c
joochlee: it's strange, in the latest commit that comment was already removed.
Are you seeing old commit maybe ??
(I have not rebased yet to identify code modification reflecting review comment easily)

api/oc_push.c
joochlee: ohh.. good point!
fixed

api/oc_push.c
joochlee: Ok, I did like this:
1. validate received receiver object list first
2. if validation fails just return false (original data is intact), if validation is success do remaining job

api/oc_push.c
```diff
@@ -0,0 +1,2605 @@
...
+						OC_PUSH_ERR("can't create receiver obj because its receiveruri is same as existing app resource (\"%.*s\")...", uri_param_len, uri_param);
+						result = OC_STATUS_FORBIDDEN;
+						goto exit;
+					}
+
+					/* create corresponding receiver object */
+					if (!_create_recv_obj(recvs_instance, rep))
+					{
+						OC_PUSH_ERR("failed to create receiver obj whose receiveruri is (\"%.*s\")...", uri_param_len, uri_param);
+						result = OC_STATUS_BAD_REQUEST;
```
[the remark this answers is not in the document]
joochlee: I'm sorry, I can't catch what you mean..

- First check (`OC_STATUS_FORBIDDEN`): if the `receiveruri` of new receiver object which will be used as the path where PUSH Update is posted could already be used by other Resource. In this case return forbidden error.
- Second check (`OC_STATUS_BAD_REQUEST`) : if creation of new receiver object fails, just return bad request error

api/oc_push.c
joochlee: right, fixed

api/oc_push.c
Danielius1922: Yeah, this one is gone, though they are still 2 TODO4ME in the code, if I'm not being deceived again.

api/oc_push.c
```diff
@@ -0,0 +1,2625 @@
...
+ */
+static bool
+_validate_recv_obj_list(oc_rep_t *obj_list)
+{
+  oc_rep_t *recv_obj, *rep;
+  bool result = false;
+  char mandatory_property_check;
+
+  if (!obj_list)
+    goto exit;
```
Danielius1922, line 1927: (This is OK, though usually `goto` is used when you need to execute some calls before returning, here just pure `return false;` would be simpler)

api/oc_push.c
Danielius1922: Yeah, that works too, good job!

api/oc_push.c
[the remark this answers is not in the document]
Danielius1922: There's a difference how the errors are handled though:

`OC_STATUS_FORBIDDEN` -> `goto exit`, so the while loop is not finished.
`OC_STATUS_BAD_REQUEST` -> result value is set, but `goto exit` is not called, thus the while loop continues

Why the difference between the errors?

api/oc_push.c
```diff
@@ -0,0 +1,2625 @@
...
+    oc_resource_bind_resource_interface(push_recv, OC_IF_RW | OC_IF_BASELINE);
+    oc_resource_set_default_interface(push_recv, OC_IF_RW);
+    oc_resource_set_discoverable(push_recv, true);
+
+    oc_resource_set_request_handler(push_recv, OC_GET, get_pushrecv, NULL);
+    oc_resource_set_request_handler(push_recv, OC_POST, post_pushrecv, NULL);
+    oc_resource_set_request_handler(push_recv, OC_DELETE, delete_pushrecv,
+                                    NULL);
+
+    oc_add_resource(push_recv);
```
Danielius1922, line 2210: We can save some instructions by moving this to line 2218:

```
    if (recvs_instance) {
      oc_add_resource(push_recv);
      recvs_instance->resource = push_recv;
      ...
    } else {
    ...
    }
```

The `oc_delete_resource` call in the else branch won't need to remove the resource from the app_resources list, which saves some instructions. :-)

api/oc_push.c
joochlee: oh yes, I know that.
I leave them intentionally as future enhancement works. 
Current implementation has already passed CTT TCs and those TODOS does not affect spec satisfaction

api/oc_push.c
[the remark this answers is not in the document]
joochlee: oops.. I see
I'm sorry, I missed `goto exit`.

fixed~!

api/oc_push.c
joochlee: agreed, fixed.

api/oc_push.c
joochlee: agreed, fixed

api/oc_push.c
```diff
@@ -0,0 +1,2628 @@
...
+
+  OC_PROCESS_END();
+}
+
+/*
+ * @brief check if any of source array is part of target array
+ * @param target
+ * @param source
+ * @return
+ * 			0: any of source is not part of target
```
Danielius1922, line 2499: this works but for `bool` it's natural to use true/false instead of 1/0

api/oc_push.c
```diff
@@ -0,0 +1,2628 @@
...
+    OC_PUSH_DBG("source or target string array is empty!");
+    return result;
+  }
+
+  for (i = 0; i < src_len; i++) {
+    for (j = 0; j < tgt_len; j++) {
+      if (!strcmp(oc_string_array_get_item(*source, i),
+                  oc_string_array_get_item(*target, j))) {
+        result = 1;
+        break;
```
Danielius1922, line 2523: You can just:

```
return true;
```

here and then you don't need the two breaks.

api/oc_push.c
joochlee: ok, done

api/oc_push.c
```diff
@@ -0,0 +1,2628 @@
...
+{
+  oc_resource_t *src_rsc;
+  char di[OC_UUID_LEN + 10];
+
+  src_rsc = (oc_resource_t *)ns_instance->user_data;
+
+  if (!ns_instance || !src_rsc) {
+    OC_PUSH_ERR("something wrong! corresponding notification selector source "
+                "resource is NULL, or updated resource is NULL!");
+    return;
```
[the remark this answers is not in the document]
joochlee: done

api/oc_push.c
joochlee: fixed

api/oc_push.c
joochlee: agree, fixed

api/oc_push.c
```diff
@@ -0,0 +1,2696 @@
...
+                  oc_string(resource->uri),
+                  oc_string(ns_instance->resource->uri));
+
+      /* resource is necessary to identify which resource is being pushed..,
+       * before sending update to target server */
+      ns_instance->user_data = resource;
+
+      /* post "event" for Resource which has just been updated */
+      if (!push_update(ns_instance)) {
+        OC_PUSH_ERR("sensing PUSH Update of \"%s\" failed!",
```
Danielius1922, line 2682: typo: sensing -> sending

Danielius1922:
Would it be possible to create an example application in apps directory? This is an extensive feature and I don't see any documentation or example, so I'm not sure people will know how to use it.

Also, I've added SonarCloud scan for this repo - it does some stylistic and linting checks, though I'm not sure the job will start for a fork. We will see when you rebase, I'd like the keep the amount of new issues reported by SC to zero.

I don't see coding problems in the pushed commits, but I haven't checked adherence to the specification... I guess there is nobody else to do it, eh?

joochlee:
> Would it be possible to create an example application in apps directory? This is an extensive feature and I don't see any documentation or example, so I'm not sure people will know how to use it.

=> 
Sure, I'll do it.
Where do you want me to add documentation? 
Add to existing `readme.md`, or create separate one ?

> 
> Also, I've added SonarCloud scan for this repo - it does some stylistic and linting checks, though I'm not sure the job will start for a fork. We will see when you rebase, I'd like the keep the amount of new issues reported by SC to zero.
> 
> I don't see coding problems in the pushed commits, but I haven't checked adherence to the specification... I guess there is nobody else to do it, eh?

=> 
I created [CR3530 - Test case for Bug2483 Bug2658](https://bugzilla.openconnectivity.org/show_bug.cgi?id=3530) and 2 CTT test case codes. 
The 2 CTT test case codes have already been merged to [ctt-files](https://github.com/openconnectivityfoundation/ctt-files/tree/BZ2483-push) repo. 
Comarch reviewed and merged them.

Danielius1922:
> => Sure, I'll do it. Where do you want me to add documentation? Add to existing `readme.md`, or create separate one ?

I don't the main readme.md to be too long, so not there. Are you creating a demo apps, if so I see that there is an .md file in the apps folder, where the demo application are described. If you want to create just documentation then I would create a new folder 'doc' and create a new .md file there which documents the feature.

joochlee:
> > => Sure, I'll do it. Where do you want me to add documentation? Add to existing `readme.md`, or create separate one ?
> 
> I don't want the main readme.md to be too long, so not there. Are you creating a demo app? If so I see that there is an .md file in the apps folder, where the demo application are described. If you want to create just documentation then I would create a new folder 'doc' and create a new .md file there which documents the feature.

I think `apps/Readme.md` is not a good place to describe additional feature such as push notification. So creating new `doc` folder is better choice.

Danielius1922:
> It looks like that `apps/Readme.md` is not a good place to describe additional feature such as push notification. So I think creating new `doc` folder is better idea.

Another alternative would be to put it in https://github.com/iotivity/iotivity.github.io . I think both options are good, pick the one you like.

joochlee:
> > It looks like that `apps/Readme.md` is not a good place to describe additional feature such as push notification. So I think creating new `doc` folder is better idea.
> 
> Another alternative would be to put it in https://github.com/iotivity/iotivity.github.io . I think both options are good, pick the one you like.

Hi Daniel, 
I pushed new commit including examples & documentations.
If you think that it looks ok, I'll rebase final commit and push it again.

port/linux/readme.md
```diff
@@ -70,4 +70,12 @@ Option and default setting in the Linux Makefile
   
 - WKCORE
 
-  enable discovery through IETF /.well-known/core on IETFs multicast ALL COAP NODES
\ No newline at end of file
+  enable discovery through IETF /.well-known/core on IETFs multicast ALL COAP NODES
+
+- PUSH 0
```
Danielius1922, line 75: Isn't this inverted? 

PUSH 1 -> enabled, PUSH 0 -> disabled

port/linux/readme.md
```diff
@@ -70,4 +70,12 @@ Option and default setting in the Linux Makefile
...
+
+  enable push notification
+
+- PUSH_DEBUG 0
```
Danielius1922, line 79: Same as above

port/linux/readme.md
joochlee: oos, that's right
fixed

apps/push_configurator_multithread_linux.c
```diff
@@ -0,0 +1,531 @@
...
+    printf("Please discovery resource first!\n");
+    return false;
+  }
+
+  return true;
+}
+
+static void cb_create_notification_selector_response(oc_client_response_t *data)
+{
+  (void) data;
```
Danielius1922, line 104: remove, `data` is used in this function

apps/push_configurator_multithread_linux.c
```diff
@@ -0,0 +1,531 @@
...
+static oc_discovery_flags_t
+cb_discovery(const char *anchor, const char *uri, oc_string_array_t types,
+                  oc_interface_mask_t iface_mask, oc_endpoint_t *endpoint,
+                  oc_resource_properties_t bm, void *user_data)
+{
+  oc_discovery_flags_t ret = OC_CONTINUE_DISCOVERY;
+
+  (void)anchor;
+  (void)iface_mask;
+  (void)bm;
```
Danielius1922, line 259: bm is used on line 274 of this function

apps/push_originserver_multithread_linux.c
```diff
@@ -0,0 +1,373 @@
...
+	oc_resource_set_pushable(res, pushable);
+	oc_resource_set_request_handler(res, OC_GET, cb_retrieve_light1, NULL);
+	/* set PUSH payload builder */
+	res->payload_builder = build_light_payload;
+	oc_add_resource(res);
+}
+
+static void signal_event_loop(void)
+{
+	pthread_mutex_lock(&mutex);
```
Danielius1922, line 210: I've changed this in master - you don't need to hold mutex when calling pthread_cond_signal

apps/push_originserver_multithread_linux.c
```diff
@@ -0,0 +1,373 @@
...
+
+	static const oc_handler_t handler =
+	{ .init = app_init, .signal_event_loop = signal_event_loop,
+			.register_resources = register_resources };
+
+#ifdef OC_STORAGE
+	oc_storage_config("./push_originserver_multithread_linux_creds");
+#endif /* OC_STORAGE */
+
+	if (pthread_mutex_init(&mutex, NULL) < 0)
```
Danielius1922, line 280: `pthread_mutex_init(&mutex, NULL) != 0`

Fixed in master already. https://linux.die.net/man/3/pthread_mutex_init - `pthread_mutex_init` returns 0 on success and error number (which are >0) on failure. (Same with `pthread_cond_init`.)

apps/docs/push.md
```diff
@@ -0,0 +1,239 @@
+# OCF Push Notification
+
+## Introduction
+OCF Push Notification lets devices send notifications to other devices without any initial request such as what observation does.
+OCF Push Notification is composed of 3 players: **Origin Server**, **Target Server**, **Configuration Client**. 
+
+![b3ca1628ad5ae3952724488e8de3593d.png](./resources/b3ca1628ad5ae3952724488e8de3593d.png)
```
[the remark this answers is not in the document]
joochlee: done

apps/docs/push.md
```diff
@@ -0,0 +1,239 @@
...
+- `push_originserver_multithread_linux`: **Origin Server**
+
+	![77fdcc907530fe864dfe719daafb5d82.png](./resources/77fdcc907530fe864dfe719daafb5d82.png)
+
+- `push_targetserver_multithread_linux`: **Target Server**
+
+	![886880151a22d9bf4ede36cc9d06e44f.png](./resources/886880151a22d9bf4ede36cc9d06e44f.png)
+
+### Onboarding
+- If you want quick test, build iotivity-lite without security (`OC_SECURITY_ENABLED=1` or `SECURE=0`).
```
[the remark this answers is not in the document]
joochlee: fixed

apps/docs/push.md
```diff
@@ -0,0 +1,239 @@
...
+	 * @param[in] resource to specify as pushable or non-pushable
+	 * @param[in] state if true the resource will be pushable if false the
+	 *                  resource will be non-pushable
+	 */
+	void oc_resource_set_pushable(oc_resource_t *resource, bool state);
+	```
+	- Call this function once when you create new pushable Resource.
+- `oc_resource_t.payload_builder`
+	```c
+	struct oc_resource_s
```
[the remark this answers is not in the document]
joochlee: done

apps/push_configurator_multithread_linux.c
joochlee: fixed

apps/push_configurator_multithread_linux.c
joochlee: fixed

apps/push_originserver_multithread_linux.c
joochlee: done

apps/push_originserver_multithread_linux.c
joochlee: I'm not clear about this. 
`mutex` is an variable defined in user application. 
how did you modify iotivity-lite codes?

Danielius1922: Just remove the `pthread_mutex_lock/unlock` lines. See: https://github.com/iotivity/iotivity-lite/commit/4be0e57b32989591bac4e5927cdb878aeebcd4bf .

joochlee: done

Danielius1922:
~format

Danielius1922:
Closed since the other PR with these commits squashed was merged (github won't delete this PR, so you can still find it if you need to see what work was done).

PR #272, opened 2022-05-27
linux: run tests from cmake instead of make

Danielius1922:


CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(OC_PKI_ENABLED ON CACHE BOOL "Enable PKI security.")
 set(OC_CLOUD_ENABLED OFF CACHE BOOL "Enable cloud communications.")
 set(OC_DEBUG_ENABLED OFF CACHE BOOL "Enable debug messages.")
 set(OC_IDD_API_ENABLED ON CACHE BOOL "Enable the Introspection Device Data API.")
-set(OC_TCP_ENABLED ON CACHE BOOL "Enable OCF communications over TCP. Necessary for Cloud communications.")
+set(OC_TCP_ENABLED OFF CACHE BOOL "Enable OCF communications over TCP. Necessary for Cloud communications.")
```
Danielius1922, line 14: There's a disrepancy between Makefile and CMake default configurations:

Makefile: 
TCP=0 (=OC_TCP is not defined)
OSCORE=1 (=OC_OSCORE is defined)

CMake: 
OC_TCP_ENABLED is true  (=OC_TCP is defined)
OC_OSCORE_ENABLED is false (=OC_OSCORE is not defined)

@WAvdBeek: Was this deliberate when @Oxymoron79 created the CMake or is it wrong? What should be the default setup regarding `TCP` and `OSCORE`?

.github/workflows/unittest.yml
```diff
@@ -1,36 +0,0 @@
...
-      - name: Checkout repository
-        uses: actions/checkout@v3
-        with:
-          submodules: "true"
-
-      # Runs a set of commands using the runners shell
-      - name: unit_test
-        run: |
-          cd port/linux
-          make test -IPV4
```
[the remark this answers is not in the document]
Danielius1922: Equivalent to the args `-DOC_IPV4_ENABLED=ON -DOC_OSCORE_ENABLED=ON` in the cmake-linux job.

WAvdBeek (approved):

CMakeLists.txt
Danielius1922: Yes, the android app is currently problematic, because it needs swig interfaces, whose generation is written in Makefile but not in CMake. But I've started looking into it last Friday.

Hmm, not sure if running all the permutations is plausible - we have at least 16 configuration switches, so that's 2^16 configurations already. :D 

What about the default values of `OC_TCP_ENABLED` and `OC_OSCORE_ENABLED` though? If the default values are not specified anywhere then I'd change the default values in CMake to match the Makefile values (=TCP disabled, OSCORE enabled).

jkralik (approved):

PR #274, opened 2022-06-03
Fix compilation and tests for static allocation

Danielius1922:


api/cloud/unittest/cloud_test.cpp
```diff
@@ -100,16 +100,16 @@ pthread_cond_t TestCloud::cv;
...
+  EXPECT_NE(nullptr, oc_cloud_get_context(0));
+  EXPECT_EQ(nullptr, oc_cloud_get_context(1));
 }
 
 TEST_F(TestCloud, cloud_status)
 {
   oc_cloud_status_t status;
   memset(&status, 0, sizeof(status));
   oc_cloud_context_t *ctx = oc_cloud_get_context(0);
-  ASSERT_NE(NULL, ctx);
```
Danielius1922, line 112: clang gives a compilation error when `NULL` instead `nullptr` is used

jkralik (approved):

PR #282, opened 2022-07-04
Conditional variable fix

Danielius1922:
Don't lock mutex before signaling conditional variable, it not
necessary.

### DIFF
--- a/api/unittest/ocapitest.cpp
+++ b/api/unittest/ocapitest.cpp
@@ -76,12 +76,6 @@ private:
   static bool s_isServerStarted;
   static bool s_isCallbackReceived;
   static oc_resource_t *s_pResource;
-#ifdef _WIN32
-  static CONDITION_VARIABLE s_cv;
-#else
-  static pthread_mutex_t s_mutex;
-  static pthread_cond_t s_cv;
-#endif
 
 public:
   static int appInit(void)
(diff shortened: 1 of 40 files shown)


api/unittest/ocapitest.cpp
```diff
@@ -106,16 +100,7 @@ class ApiHelper {
     oc_add_resource(s_pResource);
   }
 
-  static void signalEventLoop(void)
```
jkralik, line 109: this is correct?

api/unittest/ocapitest.cpp
```diff
@@ -106,16 +100,7 @@ class ApiHelper {
...
-  {
-#ifdef _WIN32
-    WakeConditionVariable(&s_cv);
-#else
-    pthread_mutex_lock(&s_mutex);
-    pthread_cond_signal(&s_cv);
-    pthread_mutex_unlock(&s_mutex);
-#endif
-  }
```
jkralik, line 118: is it correct ?
```
#ifdef _WIN32
    WakeConditionVariable(&s_cv);
#else
    pthread_cond_signal(&s_cv);
#endif
```

Danielius1922: in this file, neither the conditional variable nor the mutex were used anywhere else. probably a copy and paste from some other test.

jkralik: Ok pls check if the code is used at all. If not remove it.

Danielius1922: Other callbacks must be set and this one cannot be NULL, so there must be an empty implementation like this.

PR #285, opened 2022-07-14
Handle unauthorized code returned by endpoint

Danielius1922:


include/oc_cloud.h
```diff
@@ -43,7 +43,8 @@ typedef enum {
   OC_CLOUD_REFRESHED_TOKEN = 8,
   OC_CLOUD_LOGGED_OUT = 16,
   OC_CLOUD_FAILURE = 32,
-  OC_CLOUD_DEREGISTERED = 64
+  OC_CLOUD_DEREGISTERED = 64,
+  OC_CLOUD_UNAUTHORIZED = 128,
```
jkralik, line 47: pls changed to bit mask: eg `1 <<  7`

api/cloud/oc_cloud_manager.c
```diff
@@ -458,17 +462,22 @@ cloud_login_handler(oc_client_response_t *data)
...
-          oc_string_len(ctx->store.refresh_token) > 0) {
-        reset_delayed_callback(ctx, refresh_token, g_retry_timeout[0]);
-      }
+    goto finish;
+  }
+  if (data->code == OC_STATUS_UNAUTHORIZED) {
+    if (oc_string(ctx->store.refresh_token) &&
+        oc_string_len(ctx->store.refresh_token) > 0) {
+      ctx->store.status &= ~OC_CLOUD_LOGGED_IN;
+      ctx->store.status |= OC_CLOUD_UNAUTHORIZED;
```
jkralik, line 471: I'm not sure if this line -> because if interface goes down cloud manager is not restarted.

api/cloud/oc_cloud.c
```diff
@@ -238,13 +238,21 @@ cloud_ep_session_event_handler(const oc_endpoint_t *endpoint,
...
+  OC_DBG("[CM] cloud_ep_session_event_handler ep_state: %d\n", (int)state);
+  bool state_changed = ctx->cloud_ep_state != state;
+  if (!state_changed) {
+    return;
+  }
+
+  ctx->cloud_ep_state = state;
+  if (ctx->cloud_ep_state == OC_SESSION_DISCONNECTED && ctx->cloud_manager) {
+    bool restart = (ctx->store.status & OC_CLOUD_REGISTERED) != 0 &&
+                   (ctx->store.status & OC_CLOUD_UNAUTHORIZED) == 0;
```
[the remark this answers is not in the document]
jkralik: it seems that `OC_CLOUD_UNAUTHORIZED` is set by `login_handler` but login handler can return `OC_CLOUD_UNAUTHORIZED` -> only when refresh handler gets the error than it is issue.

fail scenario:
`start device -> send refresh token request -> encrypted response from socket -> close socket occurs -> cloud restart -> decrypted refresh_token`:
  - in this case `OC_CLOUD_UNAUTHORIZED` will be not set I think
  
 what we can do that if socket was closed -> we just delay callback of restart in `cloud_ep_session_event_handler` by `g_retry_timeout[ctx->retry_count]` -> what do you think? - or just delay callback just once - this change the order to:
`start device -> send refresh token request -> encrypted response from socket -> close socket occurs -> decrypted refresh_token  -> cloud restart`

jkralik (approved):

PR #286, opened 2022-07-18
Fix issues reported by clang-tidy

Danielius1922:


api/oc_helpers.c
```diff
@@ -288,27 +289,24 @@ oc_conv_hex_string_to_byte_array(const char *hex_str, size_t hex_str_len,
     return -1;
   }
 
-  size_t a = hex_str_len / 2.0 + 0.5;
+  long len = lround((double)hex_str_len / 2.0);
```
jkralik, line 292: is it possible to avoid double -> because operations over double is not cheap.

security/oc_pstat.c
```diff
@@ -117,9 +117,6 @@ valid_transition(size_t device, oc_dostype_t state)
       return false;
     break;
   case OC_DOS_RFPRO:
-    if (state == OC_DOS_RFOTM)
-      return false;
-    break;
```
jkralik, line 122: why this is removed ? this code seems ok

Danielius1922: It's exactly the same as the next option (`OC_DOS_RFNOP`), so clang-tidy complained about a duplicated case in the switch.

api/oc_helpers.c
Danielius1922: yes we can get rid of it

Danielius1922:
~format

jkralik (approved):

Danielius1922:
~format

PR #293, opened 2022-08-10
Update Android NDK to v25

Danielius1922:


Danielius1922:
OK, it compiles, but I have no idea if it works though. :D

jkralik (approved):

Danielius1922:
Hi @WAvdBeek do you know of somebody who could test the android port? The job was failing, because of some changes to the NDK package I think, so I've updated the NDK version and fixed compilation, but I don't have an Android device to verify it.

Danielius1922:
> Hi @WAvdBeek do you know of somebody who could test the android port? The job was failing, because of some changes to the NDK package I think, so I've updated the NDK version and fixed compilation, but I don't have an Android device to verify it.
> 
> please ask Clarke.

@cstevens can you help with this?

PR #290, opened 2022-07-26
linux: refactor tcp context and adapter

Danielius1922:

PR #304, opened 2022-08-28
add Push Notification support (CR2483)

joochlee:
add push notification feature (CR2483)


include/oc_ri.h
```diff
@@ -414,6 +432,15 @@ void oc_ri_remove_timed_event_callback(void *cb_data,
  */
 int oc_status_code(oc_status_t key);
 
+/**
+ * @brief convert the status code to string
+ *
+ * @param[in] key key the application level key of the code
+ * @return char* CoAP status code string
+ */
+/*OC_API*/
```
Danielius1922, line 441: I know how to fix this for SWIG, see: https://github.com/iotivity/iotivity-lite/pull/305/files#diff-0e60995c1ffb3932d92f32aa8b29750edc430eca97ee060bae91f74afa56d3df .

The issue is that the SWIG preprocessor doesn't include headers from other header files (in this case OC_API in oc_ri.h is defined in util/compiler.h, but the include will be ignored), so we need to either include oc_compiler.h or define OC_API in the .i file before oc_ri.h is included.

api/oc_events.h
```diff
@@ -47,6 +47,10 @@ typedef enum {
   SW_UPDATE_UPGRADING,
   SW_UPDATE_DONE,
 #endif /* OC_SOFTWARE_UPDATE */
+#if defined(OC_PUSH) && defined(OC_SERVER) && defined(OC_CLIENT) &&            \
```
Danielius1922, line 50: This is rather long, what about doing something like I did here -https://github.com/iotivity/iotivity-lite/pull/290/files#diff-99b1c8df5705e6b64ed38894790ccf9a044ff61787ec7cc55a8e4a6d57cf3745 ?

Just copy the file from my PR (I have an issue with CTT test case so I must investigate and cannot merge the PR) and add something like:

```C
#if defined(OC_PUSH) && defined(OC_SERVER) && defined(OC_CLIENT) &&   defined(OC_DYNAMIC_ALLOCATION) && defined(OC_COLLECTIONS_IF_CREATE)
#define OC_HAS_FEATURE_PUSH
#endif
```

And then you can just check for `OC_HAS_FEATURE_PUSH`.

joochlee: good Idea!

joochlee: done

.github/workflows/CLA.yml
```diff
@@ -3,10 +3,12 @@ on:
   issue_comment:
     types: [created]
   pull_request_target:
-    types: [opened,closed,synchronize]
+    types: [opened, closed, synchronize]
 
 jobs:
   CLAssistant:
+    if: ${{ false }}
```
Danielius1922, line 10: This should be reversed.

Danielius1922:
Closed since the other PR with these commits squashed was merged (github won't delete this PR, so you can still find it if you need to see what work was done).

PR #306, opened 2022-08-30
Update SonarCloud workflow

Danielius1922:
sonnar-scanner currently doesn't handle multi configuration
compilation. Thus if a file is compiled multiple times (such is
the case for files in the api folder) then only the first
compilation of such a file is analysed. So the best we can do
is enable as many features as possible and compile target in
such an order to get most of the code analysis (ie. compile
client-server target first, so we get analysis with both
OC_CLIENT and OC_SERVER defined)


joochlee:
Hi Daniel, 

This patch looks reasonable to me, so merge this patch please.
Once it is merged, I'll try my codes with this again.

Danielius1922:
> Hi Daniel,
> 
> This patch looks reasonable to me, so merge this patch please. Once it is merged, I'll try my codes with this again.

You need to give an Approve! :D Go to the Files Changed tab, click on the "Review Changes" button and select Approve.

joochlee (approved):
it looks good! :+1:

PR #311, opened 2022-09-12
Test coverage

Danielius1922:


jkralik (approved):
Is it possible to get coverage from test with `cloud server test` from plgd/hub ?

Danielius1922:
> Is it possible to get coverage from test with `cloud server test` from plgd/hub ?

Yes it should be possible. Plan for next Friday. ;)

PR #313, opened 2022-09-15
cloud: refactor unpublishing of resource links

Danielius1922:


api/cloud/oc_cloud_rd.c
```diff
@@ -208,60 +208,60 @@ publish_published_resources(void *data)
...
-  if (data->code != OC_STATUS_DELETED)
-    goto error;
-  rd_link_free(&ctx->rd_delete_resources);
-error : {
-}
 }
+#endif /* OC_DEBUG */
 
 static void
-delete_resources(oc_cloud_context_t *ctx, bool all)
```
jkralik, line 227: All need to be there - it means all resources at cloud must be unpublished
If all is not set then only selected resources need to be deleted.

Danielius1922: All was never used.

jkralik: I know but API counts with it. Maybe put there comment about that device can unpublish all resources without ins

Danielius1922: API? But it's a static helper function, surely that cannot be regarded as an API function.

api/cloud/oc_cloud_rd.c
```diff
@@ -208,60 +208,60 @@ publish_published_resources(void *data)
...
     return;
   }
-  if (ctx->rd_delete_resources) {
-    rd_delete(ctx->cloud_ep, ctx->rd_delete_resources, ctx->device,
-              delete_resources_handler, LOW_QOS, ctx);
+
+#ifdef OC_DEBUG
+  oc_response_handler_t handler = delete_resources_handler;
+#else  /* !OC_DEBUG */
+  oc_response_handler_t handler = NULL;
```
jkralik, line 240: `handler` need to be always set. otherwise segfault can occurs.

jkralik (approved):

PR #326, opened 2022-10-13
cloud: update configuration resource handling

Danielius1922:
Update required state and payload condition for the update to go through.

### DIFF
--- a/api/cloud/oc_cloud_resource.c
+++ b/api/cloud/oc_cloud_resource.c
@@ -106,44 +106,26 @@ get_cloud(oc_request_t *request, oc_interface_mask_t interface, void *user_data)
 static bool
 cloud_update_from_request(oc_cloud_context_t *ctx, oc_request_t *request)
 {
-  uint8_t required = 0;
-  bool changed = false;
   cloud_conf_update_t data;
   memset(&data, 0, sizeof(data));
 
-  if (oc_rep_get_string(request->request_payload, OC_RSRVD_ACCESSTOKEN,
-                        &data.access_token, &data.access_token_len)) {
-    required |= 0x01;
-    changed = true;
-  }
+  bool has_at =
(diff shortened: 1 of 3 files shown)


api/cloud/oc_cloud_resource.c
```diff
@@ -161,13 +145,13 @@ post_cloud(oc_request_t *request, oc_interface_mask_t interface,
   bool request_invalid_in_state = true;
   switch (ctx->store.cps) {
   case OC_CPS_UNINITIALIZED:
-  case OC_CPS_READYTOREGISTER:
-  case OC_CPS_FAILED:
-  case OC_CPS_DEREGISTERING:
     request_invalid_in_state = false;
     break;
+  case OC_CPS_READYTOREGISTER:
```
Danielius1922, line 150: @jkralik I think this might be the opposite of what we wanted, certainly the hub tests don't like it.

jkralik: why ? if we moved it to line 145 it will be fixed ?

Danielius1922: After this change in `OC_CPS_READYTOREGISTER`, `OC_CPS_DEREGISTERING`, and  `OC_CPS_FAILED` state you can only set empty "cis", for every other update the `request_invalid_in_state` will be true and thus line 165 will return OC_STATUS_BAD_REQUEST. Previously, in these states you could do all kinds of updates. It seems that at least the hub tests require the device to be configurable in the `OC_CPS_READYTOREGISTER` state. Or do we want to rewrite tests in plgd repos?

jkralik: Oh ok. For these you can allow too configure it with all 4 values.

jkralik (approved):

PR #342, opened 2022-11-22
oc_reset_v1: allow to set close_all_tls_connections_immediately

jkralik:


jkralik:
~format

Danielius1922:
@jkralik what about a unit test for the new functions? You should be able to mock a connection using `oc_tls_add_peer` and then the reset should remove it.

messaging/coap/engine.c
```diff
@@ -924,14 +910,6 @@ coap_receive(oc_message_t *msg)
     }
   }
 
-#ifdef OC_SECURITY
-  if (coap_status_code == CLOSE_ALL_TLS_SESSIONS) {
```
Danielius1922, line 928: This was redundant?

jkralik: Yes. it was not obvious when the tls connections will be closed. - refactored

jkralik:
~format

Danielius1922 (approved):

PR #343, opened 2022-11-25
oc_resource_set_access_in_RFTOM: allow access to the resource in the RFOTM state

jkralik:
The customer has a loT of devices in rack and wants to know which one he's going to manipulate, so he wants to turn on the light before transferring ownership.


jkralik:
~format

jkralik:
~format

jkralik:
~format

api/oc_discovery.c
```diff
@@ -379,9 +403,21 @@ filter_oic_1_1_resource(oc_resource_t *resource, oc_request_t *request,
...
+#ifdef OC_SECURITY
+  bool owned_for_SVRs =
+    (oc_core_is_SVR(resource, device_index) &&
+     (((oc_sec_get_pstat(device_index))->s != OC_DOS_RFOTM) ||
+      oc_tls_num_peers(device_index) != 0));
+#else  /* OC_SECURITY */
+  bool owned_for_SVRs = false;
+#endif /* OC_SECURITY */
+
   while (eps != NULL) {
```
Danielius1922, line 416: What about switching while cycle to for, so the assignment of next is writen just once:
```
for(; eps!=NULL; eps = eps->next)
```

api/oc_collection.c
```diff
@@ -807,18 +800,10 @@ oc_handle_collection_linked_list_request(oc_request_t *request)
...
-         * Also, exclude all endpoints that are not associated with the
-         * interface through which this request arrived. This is achieved by
-         * checking if the interface index matches.
-         */
-        if (((link->resource->properties & OC_SECURE) &&
-             !(eps->flags & SECURED)) ||
-            (request->origin && request->origin->interface_index != -1 &&
-             request->origin->interface_index != eps->interface_index)) {
-          goto next_eps;
+        if (oc_filter_out_ep_for_resource(eps, link->resource, request->origin,
```
Danielius1922, line 803: The `oc_filter_out_ep_for_resource` function features this additional code

```C
  if (request_origin &&
      (((request_origin->flags & IPV4) && (ep->flags & IPV6)) ||
       ((request_origin->flags & IPV6) && (ep->flags & IPV4)))) {
    return true;
  }
```

which is not in the original code. Is that OK?

jkralik: It is :) it's from discovery but not used and we want to use it :)

jkralik:
~format

Danielius1922 (approved):

PR #348, opened 2022-12-21
Fixes of compilations warning

SiMet:


SiMet:
~format

Danielius1922 (approved):

Danielius1922:
@SiMet Could you please check why the CTT job doesn't seem to start? However, these changes are minor so I can force merge this if you cannot wait.

SiMet:
> @SiMet Could you please check why the CTT job doesn't seem to start? However, these changes are minor so I can force merge this if you cannot wait.

@Danielius1922 I fixed issue on jenkins and run it once again.

Danielius1922:
> @Danielius1922 I fixed issue on jenkins and run it once again.

@SiMet: Thanks, indeed it's back to the state when just CT3.1.9, CT1.7.9.6.2 and CT1.7.9.6.3 cases are failing, so I'll merge this.

BTW will those test cases be fixed? If not then disabling them would make sense. Right now we always have a failure and then I check if the failures are just those three, in which case I ignore it and merge the PR anyway.

SiMet:
> > @Danielius1922 I fixed issue on jenkins and run it once again.
> 
> @SiMet: Thanks, indeed it's back to the state when just CT3.1.9, CT1.7.9.6.2 and CT1.7.9.6.3 cases are failing, so I'll merge this.
> 
> BTW will those test cases be fixed? If not then disabling them would make sense. Right now we always have a failure and then I check if the failures are just those three, in which case I ignore it and merge the PR anyway.

We had mail exchange what should be done here. Eonti certificates which where required for this 3 testcases could not be generated. So test setup is not meet.

As I remmember we could dissable testcases 1.7.9.6.2 and 1.7.9.6.3 in PICS but 3.1.9 which tests two certs provider could not be dissabled

Danielius1922:
@SiMet, yes I recall the conversation, it was back in November. you said that we are waiting for OCF. Is there anything new there? If a solution is still far away (several months), can't we come up with some sort of a quick solution in the wrapper .py script? Something like ignoring the failure if the failed test cases are those three -  1.7.9.6.2, 1.7.9.6.3 and 3.1.9?

PR #356, opened 2023-01-20
Link to system mbedtls

Oxymoron79:
This PR adds a `BUILD_MBEDTLS` CMake option:

- If set to `ON` (default) the mbedtls library is patched and built by the iotivity-lite build system.
- If set to `OFF`, the iotivity-lite build system searches for the mbedtls library using CMake's `find_package` function and verifies that the library has been built with the OCF patches applied.

To test this feature, I used patched, built and installed the mbedtls library from the `deps/mbedtls` submodule before building iotivity-lite:
To patch mbedtls I used the `deps/mbedtls-patch.cmake` script:
```
$ cmake -DOC_DYNAMIC_ALLOCATION_ENABLED=ON -DOC_PKI_ENABLED=ON -P deps/mbedtls-patch.cmake
```
Then I configured, built and installed mbedtls as a shared library:
```
$ cd deps/mbedtls
$ cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PROGRAMS=OFF -DENABLE_TESTING=OFF -DUSE_SHARED_MBEDTLS_LIBRARY=ON .
$ make install
```

After that, with `BUILD_MBEDTLS=OFF` iotivity-lite was able to 

- find the mbedtls shared library with the `find_package` function
- verify that mbedtls was patched by checking for the `mbedtls_x509write_crt_set_subject_alt_names` function
- compile iotivity-lite and link to the mbedtls shared library

But to make the patched mbedtls library compile successfully, I had to modify the OCF patches and remove the generic entropy pool setting, since that requires parts of the iotivity-lite sources.

I am not sure what impact this modification of the OCF patches for mbedtls has and whether that would be acceptable.
I also haven't tested whether `BUILD_MBEDTLS=OFF` works on all platforms and with different build configurations.


Danielius1922:
Nice job, Martin!

We've talked with @jkralik and if we understand the problem with the `MBEDTLS_ENTROPY_C` correct, the issue is that `mbedtls_platform_entropy_poll` is added by the patch. Which includes `port/oc_random.h` and uses `oc_random_value` from iotivity-lite. So we decided that instead of directly using `oc_random_value` I will update the patch and add a function pointer and a setter for this pointer to mbedtls and this pointer will be used instead of `oc_random_value`. Iotivity lite will set this function to `oc_random_value()` when it initializes mbedtls.
Similar issue might be with `MBEDTLS_PLATFORM_STD_EXIT` which is defined as `oc_exit`, but we will solve it the same way. And this should allow us to keep `MBEDTLS_ENTROPY_C` enabled. I'll add commits to this PR if that's OK.

Also to test the various compilation options I'm thinking about modifying the GH workflows. In most workflows we have `workflow_dispatch` enabled, so one can run a workflow manually. And dispatching can be parameterized. I'll add a bool, which controls whether preinstalled or submodule mbedtls should be used. The workflows will need to be modified to add an mbedtls installation step, but that shouldn't be complicated.

jkralik (approved):

PR #358, opened 2023-01-20
Update github workflows

Danielius1922:
Make tests that use CMake to build configurable, so they can be built in such a way that they preinstall mbedTLS and then use the preinstalled libraries.


CMakeLists.txt
```diff
@@ -380,7 +380,7 @@ if(OC_SECURITY_ENABLED)
         get_target_property(mbedtls_INTERFACE_LINK_LIBRARIES MbedTLS::mbedtls INTERFACE_LINK_LIBRARIES)
         set(CMAKE_REQUIRED_INCLUDES ${mbedtls_INTERFACE_INCLUDE_DIRECTORIES})
         set(CMAKE_REQUIRED_LIBRARIES ${mbedtls_INTERFACE_LINK_LIBRARIES})
-        check_symbol_exists(mbedtls_x509write_crt_set_subject_alt_names "mbedtls/x509_crt.h" MBEDTLS_PATCHED)
+        check_symbol_exists(MBEDTLS_KEY_EXCHANGE_ECDH_ANON_ENABLED "mbedtls/mbedtls_config.h" MBEDTLS_PATCHED)
```
Danielius1922, line 383: @Oxymoron79, I had to change the check here, because the original didn't work for all compilation options. `MBEDTLS_KEY_EXCHANGE_ECDH_ANON_ENABLED` is added always and is not `ifdef`-ed by anything, so it works reliably.

PR #360, opened 2023-01-20
cloud: Fix heap-use-after-free when deregisting from cloud

Danielius1922:
Setting empty server forces deregistration from the original cloud server (if the device is connected to it). This is done by sending a deregister request to the cloud.
When the deregistration is triggered by a receiving a POST request form the server then creating a new request ovewrites the global shared buffer for requests. Because the original POST request is waiting for response and will try to use the buffer already used and freed by the deregistration request.

The solution is to execute deregistration only after the POST response has been sent.

### DIFF
--- a/api/cloud/oc_cloud.c
+++ b/api/cloud/oc_cloud.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2019 Intel Corporation
  * Copyright 2019 Jozef Kralik All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the "License"),
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
@@ -24,6 +24,7 @@
 #include "api/oc_server_api_internal.h"
 #include "oc_api.h"
 #include "oc_cloud_context_internal.h"
+#include "oc_cloud_deregister_internal.h"
 #include "oc_cloud_internal.h"
 #include "oc_cloud_manager_internal.h"
 #include "oc_cloud_store_internal.h"
@@ -35,6 +36,14 @@
 #ifdef OC_SECURITY
 #include "security/oc_tls.h"
 #endif /* OC_SECURITY */
+
+void
+cloud_reset_delayed_callback(void *cb_data, oc_trigger_t callback,
+                             uint16_t seconds)
+{
+  oc_remove_delayed_callback(cb_data, callback);
+  oc_set_delayed_callback(cb_data, callback, seconds);
+}
 
 bool
 cloud_is_connection_error_code(oc_status_t code)
@@ -87,6 +96,7 @@ static void
 cloud_manager_restart(oc_cloud_context_t *ctx)
 {
   cloud_manager_stop(ctx);
+  cloud_deregister_stop(ctx);
   oc_remove_delayed_callback(ctx, start_manager);
   oc_set_delayed_callback(ctx, start_manager, 0);
 }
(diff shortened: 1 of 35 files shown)


api/cloud/oc_cloud.c
```diff
@@ -106,48 +115,119 @@ cloud_close_endpoint(oc_endpoint_t *cloud_ep)
...
+  }
+
+  cloud_api_param_t *p = alloc_api_param();
+  if (p == NULL) {
+    OC_ERR("[Cloud] cannot allocate cloud parameters for reset");
+    return false;
+  }
+  p->ctx = ctx;
+  p->timeout = timeout;
+  cloud_reset_delayed_callback(p, cloud_deregister_on_reset_async, 0);
```
jkralik, line 208: this doesn't remove `cloud_deregister_on_reset_async` from delayed callbacks -> it always add new one.

api/cloud/oc_cloud_deregister.c
```diff
@@ -0,0 +1,242 @@
...
+    OC_ERR("Failed to refresh access token for deregister");
+    cloud_api_param_t *p = (cloud_api_param_t *)data;
+    free_api_param(p);
+    cloud_context_clear(ctx);
+    return;
+  }
+
+  // invoke in a delayed callback so, the cloud_api_param_t* structure allocated
+  // by cloud_deregister is deallocated before new one is allocated
+  cloud_reset_delayed_callback(data, cloud_deregister_refreshed_token_async, 0);
```
jkralik, line 171: this always add `cloud_deregister_refreshed_token_async` because data is always allocated

api/cloud/oc_cloud_apis.c
```diff
@@ -38,18 +39,20 @@
 
 #include <assert.h>
 
-OC_MEMB(api_params, cloud_api_param_t, 1);
+OC_MEMB(g_api_params, cloud_api_param_t,
+        2); // cloud_deregister might invoke cloud_refresh_token or cloud_login
```
jkralik, line 43: you need to multiply it by number of devices

api/cloud/oc_cloud_deregister.c
```diff
@@ -0,0 +1,351 @@
...
+  }
+  return 0;
+}
+
+static bool
+cloud_match_context(const void *cb_data, const void *filter_data)
+{
+  const cloud_api_param_t *p = (const cloud_api_param_t *)cb_data;
+  const oc_cloud_context_t *ctx = (const oc_cloud_context_t *)filter_data;
+  return p->ctx == ctx;
```
jkralik, line 336: you need to to free(p) if this condition true - otherwise it leak memory

api/cloud/oc_cloud_context.c
```diff
@@ -110,6 +111,7 @@ cloud_context_clear(oc_cloud_context_t *ctx)
   memset(ctx->cloud_ep, 0, sizeof(oc_endpoint_t));
   ctx->cloud_ep_state = OC_SESSION_DISCONNECTED;
   cloud_manager_stop(ctx);
+  cloud_deregister_stop(ctx);
```
jkralik, line 114: And what about when  someone set new cloud during registration? Do we have covered this usecase?
I think we need add somewhere this `cloud_deregister_stop`.

Danielius1922: Should be handled now, I've added `cloud_deregister_stop` after each call of `cloud_close_endpoint` (which timeouts all request handlers) in `api/cloud` files.

jkralik (approved):

PR #367, opened 2023-02-01
Add BUILD_TINYCBOR CMake option

Danielius1922:
If set to ON (default) the tinyCBOR library built by the iotivity-lite build system.
If set to OFF, the iotivity-lite build system searches for the tinyCBOR library using CMake's find_package.


CMakeLists.txt
```diff
@@ -21,6 +21,10 @@ endif()
 ######## Build configuration options ########
 set(BUILD_EXAMPLE_APPLICATIONS ON CACHE BOOL "Build example applications.")
 set(BUILD_MBEDTLS ON CACHE BOOL "Build Mbed TLS library. When set to OFF, the Mbed TLS library with the OCF patches has to be provided.")
+set(OC_INSTALL_MBEDTLS ON CACHE BOOL "Include Mbed TLS in installation")
```
jkralik, line 24: when I use this flags ?

Danielius1922: It changes the behavior of `make install` (or `cmake --install`). When `OC_INSTALL_TINYCBOR` and `OC_INSTALL_MBEDTLS` is OFF then only iotivity-lite targets are installed. When they are on then also tinycbor and mbedtls targets are installed.

jkralik (commented):
Pls add one test with `-DBUILD_TINYCBOR=OFF -DBUILD_MBEDTLS=OFF `

jkralik (approved):

Danielius1922:
@SiMet I seem to have broken something with the CTT job with this merge. It seems to me from the log that the binaries didn't start, which suggests either a broken build or a runtime error. Do you please have any logs? And if it is a build problem, could you describe to me how do you build the binaries? I would add a compilation job with that configuration, so this problem doesn't repeat.

PR #371, opened 2023-02-06
Fix for build break - !370

SiMet:


Danielius1922 (approved):

jkralik (approved):

Danielius1922:
@SiMet thank you for your help!

As a side issue - given that you use port/linux would it be possible to switch to CMake at some point in the future? Eventually, we would like to switch to just a single build system so we don't have to maintain two. (Though right know we cannot remove it anyway, because CMake doesn't support Swig build with Java bindings; but I'll get to that eventually and then I'll try to get rid of the Makefiles).

The equivalent commands should be:

```shell
mkdir build && cd build
cmake .. -DOC_TCP_ENABLED=ON -DOC_MNT_ENABLED=ON -DOC_SOFTWARE_UPDATE_ENABLED=ON -DOC_CLOUD_ENABLED=ON -DOC_IPV4_ENABLED=OFF -DOC_COLLECTIONS_IF_CREATE_ENABLED=ON -DOC_OSCORE_ENABLED=ON
cmake --build .
```

And the binaries will be in the build folder.

PR #379, opened 2023-02-10
cloud: configure keepalive

jkralik:
Enable Setting of next ping Delay and timeout through oc_cloud_set_keepalive callback Function.

### DIFF
--- a/api/cloud/oc_cloud.c
+++ b/api/cloud/oc_cloud.c
@@ -50,7 +50,7 @@ cloud_reset_delayed_callback_ms(void *cb_data, oc_trigger_t callback,
                                 uint64_t milliseconds)
 {
   oc_remove_delayed_callback(cb_data, callback);
-  oc_set_delayed_callback_ms(cb_data, callback, milliseconds);
+  oc_set_delayed_callback_ms_v1(cb_data, callback, milliseconds);
 }
 
 bool
--- a/api/cloud/oc_cloud.c
+++ b/api/cloud/oc_cloud.c
@@ -45,6 +45,14 @@ cloud_reset_delayed_callback(void *cb_data, oc_trigger_t callback,
   oc_set_delayed_callback(cb_data, callback, seconds);
 }
 
+void
+cloud_reset_delayed_callback_ms(void *cb_data, oc_trigger_t callback,
+                                uint64_t milliseconds)
+{
+  oc_remove_delayed_callback(cb_data, callback);
+  oc_set_delayed_callback_ms(cb_data, callback, milliseconds);
+}
+
 bool
 cloud_is_connection_error_code(oc_status_t code)
 {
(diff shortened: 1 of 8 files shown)


api/cloud/oc_cloud.c
```diff
@@ -45,6 +45,14 @@ cloud_reset_delayed_callback(void *cb_data, oc_trigger_t callback,
   oc_set_delayed_callback(cb_data, callback, seconds);
 }
 
+void
+cloud_reset_delayed_callback_ms(void *cb_data, oc_trigger_t callback,
+                                uint64_t milliseconds)
+{
+  oc_remove_delayed_callback(cb_data, callback);
+  oc_set_delayed_callback_ms(cb_data, callback, milliseconds);
```
Danielius1922, line 53: `cloud_reset_delayed_callback_ms` and oc_set_delayed_callback_ms should have the same width of the interval argument. However,`oc_set_delayed_callback_ms` takes `uint16_t` interval -> that's max 65536, so 65 seconds... that's really not much. We should change the API to at least uint32_t or uint64_t.

jkralik: `oc_set_delayed_callback_ms_v1` deprecates `oc_set_delayed_callback_ms`

jkralik:
~format

include/oc_cloud.h
```diff
@@ -115,6 +133,15 @@ typedef struct oc_cloud_context_t
   int selected_identity_cred_id; /**< Selected identity cert chain. -1(default)
                                     means any*/
   bool cloud_manager;
+
+  oc_cloud_on_keepalive_response_cb_t on_keepalive_response; /**< Callback
+                                                               invoked on
+                                                               keepalive
+                                                               response */
+  uint16_t keepalive_ping_timeout; /**< Timeout for keepalive ping in seconds */
+  void *keepalive_user_data; /**< User data provided to the keepalive response
```
Danielius1922, line 142: Sonar will start complaining if a struct has more than 20 members (this now has 20, so it's now as big as it can be).

I would put these three in a substruct to associate them together and keep the number below 20, something like:

```C
typedef oc_cloud_keepalive_t {
  oc_cloud_on_keepalive_response_cb_t on_response; /**< Callback invoked on keepalive response */
  void* user_data; // /**< User data provided to the keepalive response callback */
  uint16_t ping_timeout; /**< Timeout for keepalive ping in seconds */
} oc_cloud_keepalive_t;
```

Also maybe we should keep the `cloud_manager` last, `sizeof(bool)==1` and one byte members tend to mess up padding if they are in the middle of the struct. (There are `uint8_t` members which have the same issue, but since this is a public struct, I guess we shouldn't mess with it too much).

jkralik: Fixed

jkralik: Fixed

Danielius1922 (approved):

PR #383, opened 2023-02-16
security: enable MBEDTLS_MD_SHA384 and MBEDTLS_ECP_DP_SECP384R1

Danielius1922:
Allow certificates and CSRs to be signed with SHA384. Use SHA384 as the default when generating certificates and CSRs.


security/oc_certs.c
```diff
@@ -42,6 +42,63 @@
 #define CN_UUID_PREFIX "CN=uuid:"
 #define CN_UUID_PREFIX_LEN (sizeof(CN_UUID_PREFIX) - 1)
 
+// allowed message digests signature algorithms
+static int g_allowed_mds_mask = MBEDTLS_X509_ID_FLAG(MBEDTLS_MD_SHA256) |
+                                MBEDTLS_X509_ID_FLAG(MBEDTLS_MD_SHA384);
+
+// message digest used for signature of generated certificates or certificate
+// signing requests
+static mbedtls_md_type_t g_signature_md = MBEDTLS_MD_SHA384;
```
jkralik, line 51: Default SHA256

security/oc_certs.c
```diff
@@ -42,6 +42,63 @@
...
+oc_certs_set_signature_md_algorithm(mbedtls_md_type_t md)
+{
+  g_signature_md = md;
+  OC_DBG("signature message digest: %d", (int)g_signature_md);
+}
+
+void
+oc_certs_md_algorithm_allow(mbedtls_md_type_t md)
+{
+  g_allowed_mds_mask |= MBEDTLS_X509_ID_FLAG(md);
```
jkralik, line 69: Only sha256 and sha384 need to be allowed

security/oc_certs_internal.h
```diff
@@ -216,6 +216,70 @@ int oc_certs_parse_role_certificate(const unsigned char *rcert,
...
+
+/**
+ * @brief Globally allow a single given message digest to be used in signatures.
+ *
+ * @param md message digest to allow
+ */
+void oc_certs_md_algorithm_allow(mbedtls_md_type_t md);
+
+/**
+ * @brief Globally allow a single given message digest to be used in signatures.
```
jkralik, line 243: Disallow

security/oc_certs.c
jkralik: It seems fine. Just update this to sha256 and fix build for esp32.

security/oc_keypair_internal.h
```diff
@@ -0,0 +1,141 @@
...
+
+#include <mbedtls/ecp.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define OC_ECDSA_PUBKEY_SIZE (91)
```
jkralik, line 34: Is this size enough large for p384 (pubkey_size na privkey_size) i thnik this size depends on size of curve.

security/oc_csr_internal.h
```diff
@@ -65,10 +86,10 @@ bool oc_sec_csr_verify_signature(mbedtls_x509_csr *csr, int md_flags);
  * @return 0 on success
  * @return -1 on error
  */
-int oc_sec_csr_validate(const unsigned char *csr, size_t csr_len,
-                        mbedtls_pk_type_t pk_type, int md_flags,
-                        oc_string_t *subject_DN, uint8_t *public_key,
-                        size_t public_key_size);
+int oc_sec_csr_validate2(const unsigned char *csr, size_t csr_len,
```
jkralik, line 89: Does this deprecate function `oc_sec_csr_validate`?

security/oc_keypair.c
```diff
@@ -44,86 +47,135 @@ oc_sec_get_ecdsa_keypair(size_t device)
...
-    if (!kp) {
-      return false;
+#define PROP_PUBLIC_KEY "public_key"
+#define PROP_PRIVATE_KEY "private_key"
+#define STR_LEN(x) (sizeof(x) - 1)
+
+  const uint8_t *public_key = NULL;
+  const uint8_t *private_key = NULL;
+  uint8_t private_key_size = 0;
+  uint8_t public_key_size = 0;
```
jkralik, line 89: for p521 it is big enough?

jkralik: maybe `size_t` will be better.

security/oc_csr_internal.h
jkralik: if yes then we need to use `oc_sec_csr_validate_v1` because we don't want to introduce a new convention.

Danielius1922: You're too quick! I've pushed this just to get analysis from Sonar, this function will be removed once I rewrite the last test.

security/oc_keypair.c
Danielius1922: With `MBEDTLS_ECP_DP_SECP521R1_ENABLED` defined -> private key has size=223 and public key has size=158. So it would fit, but it's assigned to a `size_t` member of the keypair struct, so no there's no reason not use `size_t` anyway.

jkralik (approved):

Issue #385, opened 2023-02-21
Unable to connect to simple server using TCP due to OC_HAS_FEATURE_TCP_ASYNC_CONNECT condition


TomCahill:
The simpleserver example doesn't have OC_CLIENT defined against it which results in a FALSE condition wrapping `OC_HAS_FEATURE_TCP_ASYNC_CONNECT`.
https://github.com/iotivity/iotivity-lite/blob/fbb13f0f0a52e4aa96f9d1a7d643885601b753dd/util/oc_features.h#L23-L27

This in turn means that TCP isn't handled when built with it:
`cmake ../iotivity-lite -DCMAKE_BUILD_TYPE=Debug -DOC_DEBUG_ENABLED=1 -DOC_TCP_ENABLED=1`

Danielius1922:
Connect shouldn't be available without OC_CLIENT, at least that's the way I understand it. I suspect that I might've incorrectly `#ifdef`-ed something with `OC_HAS_FEATURE_TCP_ASYNC_CONNECT` when I reworked the connect implementation to be asynchronous.

I think this - https://github.com/iotivity/iotivity-lite/blame/master/port/linux/ipadapter.c#L1169 - might be the problematic line. It is under the async connect ifdef, but previously was under the `OC_TCP`. I'll create a PR and change the ifdefs, could you please test it? 

Thank you, for contributing!

Danielius1922:
@TomCahill connecting to simpleserver and receiving a message should now work. See: https://github.com/iotivity/iotivity-lite/pull/387 .

Danielius1922:
Possibly another bug, it seems that we haven't tested the server only and client only APIs after the connection rework. Sorry about that. 

The point of interest in the new log seems to be the "connection was closed" message. That is very suspicious (unless your client closed the connection).

Danielius1922:
@TomCahill I've pushed another fix onto the branch.

TomCahill:
@Danielius1922 Perfecto, just tested it with my setup and it's working as expected, thank you for spending time looking into it.

Danielius1922:
> @Danielius1922 Perfecto, just tested it with my setup and it's working as expected, thank you for spending time looking into it.

Thanks for your help too, there's one less bug in the library!

PR #386, opened 2023-02-22
Fixed the condition around the OC_HAS_FEATURE_TCP_ASYNC_CONNECT flag preventing handling of TCP

TomCahill:
Relates to: #385

The flag was only active when OC_CLIENT was passed which breaks simpleserver which only has OC_SERVER passed to it.

I've updated the condition to include OC_SERVER which covers the case for simpleserver and some others I expect.


TomCahill:

PR #390, opened 2023-02-27
cloud: reinitialize cloud storage for shut down during deregistering

jkralik:
In the case of a factory reset, the cloud data may remain on the device when the device is shut down during de-registration.


api/cloud/oc_cloud_context.c
```diff
@@ -38,6 +38,24 @@
 OC_LIST(g_cloud_context_list);
 OC_MEMB(g_cloud_context_pool, oc_cloud_context_t, OC_MAX_NUM_DEVICES);
 
+static void
+reinitialize_cloud_storage(oc_cloud_context_t *ctx)
+{
+#ifdef OC_SECURITY
+  const oc_sec_pstat_t *ps = oc_sec_get_pstat(ctx->device);
```
Danielius1922, line 45:
```C
#ifdef OC_SECURITY
  const oc_sec_pstat_t *ps = oc_sec_get_pstat(ctx->device);
  if (ps->s != OC_DOS_RFOTM && ps->s != OC_DOS_RESET) {
    return;
  }
#endif /* OC_SECURITY  */

if (cloud_is_deregistering(ctx)) {
...
```

jkralik: fixed

Danielius1922 (approved):

PR #394, opened 2023-03-02
cloud server docker images print time for each line log

jkralik:


port/oc_log.h
```diff
@@ -42,6 +42,8 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <sys/time.h>
```
Danielius1922, line 45: I think `<sys/*>` headers are Unix only.

Maybe `port/oc_clock.h` or `oc_clock_util.h` might have some portable functions.

jkralik: removed

CMakeLists.txt
```diff
@@ -797,6 +797,9 @@ if(BUILD_TESTING AND(UNIX OR MINGW))
     endmacro()
 
     file(GLOB COMMONTEST_SRC tests/gtest/*.cpp)
+    if (OC_DEBUG_ENABLED)
+        list(APPEND COMMONTEST_SRC ${PROJECT_SOURCE_DIR}/api/oc_clock.c)
```
Danielius1922, line 801: This is not needed, no?

jkralik: no it is not needed - removed

Danielius1922 (approved):

jkralik:
@SiMet pls could you check `ocf conformance testing`? It seems to be hanged.

PR #398, opened 2023-03-10
client: return OC_CONNECTION_CLOSED to client handler

jkralik:
when connection has been closed by peer or client.

### DIFF
--- a/api/oc_ri.c
+++ b/api/oc_ri.c
@@ -1633,11 +1633,11 @@ oc_ri_remove_client_cb(void *data)
 }
 
 static void
-notify_client_cb_timeout(oc_client_cb_t *cb)
+notify_client_cb_with_code(oc_client_cb_t *cb, oc_status_t code)
 {
-  OC_DBG(
-    "notify_client_cb_timeout - calling handler with request timeout for %d %s",
-    cb->method, oc_string(cb->uri));
+  OC_DBG("notify_client_cb_with_code - calling handler with request timeout "
+         "for %d %s",
+         cb->method, oc_string(cb->uri));
   oc_ri_remove_timed_event_callback(cb, &oc_ri_remove_client_cb);
   oc_ri_remove_timed_event_callback(
     cb, &oc_ri_remove_client_cb_with_notify_timeout_async);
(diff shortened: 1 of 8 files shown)


api/oc_ri.c
```diff
@@ -1680,7 +1680,7 @@ oc_ri_free_client_cbs_by_mid(uint16_t mid)
     if (!cb->multicast && !cb->discovery && cb->ref_count == 0 &&
         cb->mid == mid) {
       cb->ref_count = 1;
-      notify_client_cb_timeout(cb);
+      notify_client_cb_with_code(cb, OC_REQUEST_TIMEOUT);
```
Danielius1922, line 1683: I think this also shouldn't use  the `OC_REQUEST_TIMEOUT` code.

api/oc_ri.c
```diff
@@ -1697,7 +1697,7 @@ oc_ri_free_client_cbs_by_endpoint(const oc_endpoint_t *endpoint)
     if (!cb->multicast && !cb->discovery && cb->ref_count == 0 &&
         oc_endpoint_compare(&cb->endpoint, endpoint) == 0) {
       cb->ref_count = 1;
-      notify_client_cb_timeout(cb);
+      notify_client_cb_with_code(cb, OC_CONNECTION_CLOSED);
```
Danielius1922, line 1700: Both `oc_ri_free_client_cbs_by_mid` and `oc_ri_free_client_cbs_by_endpoint` are public API, so they can be called whenever, not just when connection is closed. So I think `OC_CONNECTION_CLOSED` doesn't always correctly describe what's happening. I guess both functions cancel the callback, so maybe something like `OC_CANCELLED` ? 
Or maybe we should add a status code parameter, so they caller can determine which code should be received in the cancelled callback. Although that would be an API change.

jkralik: fixed

api/oc_ri.c
jkralik: fixed

api/oc_ri.c
```diff
@@ -1689,22 +1689,35 @@ oc_ri_free_client_cbs_by_mid(uint16_t mid)
 }
 
 void
-oc_ri_free_client_cbs_by_endpoint(const oc_endpoint_t *endpoint)
+oc_ri_free_client_cbs_by_mid(uint16_t mid)
+{
+  oc_ri_free_client_cbs_by_mid_v1(mid, OC_REQUEST_TIMEOUT);
```
Danielius1922, line 1694: Wouldn't `OC_CANCELLED` be a better default than a timeout? (I've already changed the value from `OC_STATUS_SERVICE_UNAVAILABLE` to `OC_REQUEST_TIMEOUT`, so I don't think changing the code value is a problem).

jkralik: Ok. I didn't know it. Fixed

Danielius1922 (approved):

PR #403, opened 2023-03-17
cloud: expose oc_cloud_access_* functions to public API

jkralik:


include/oc_cloud_access.h
```diff
@@ -0,0 +1,119 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Common configuration arguments for cloud_access functions
+typedef struct oc_cloud_access_conf_t
+{
+  oc_endpoint_t *endpoint;       ///< cloud endpoint
```
Danielius1922, line 37: const?

jkralik: fixed

PR #407, opened 2023-03-24
Updates for plgd/time

Danielius1922:


util/oc_macros.h
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ******************************************************************/
+
+#ifndef OC_MACROS_H
+#define OC_MACROS_H
+
+#define OC_TO_STR(x) #x
+#define OC_STRLEN(x) (sizeof(x) - 1)
```
jkralik, line 23: this name can lead to issue. If someone put there pointer then he will gets sizeof pointer....

Danielius1922: I think the static analyzer should catch that case, or can we think of a better name?

jkralik: or `#define OC_CHAR_ARRAY_LEN(x) (sizeof(x) - 1)`

jkralik (approved):

PR #408, opened 2023-03-28
PlgdTime: Fetch time from endpoint

Danielius1922:


include/plgd/plgd_time.h
```diff
@@ -153,8 +165,96 @@ bool plgd_time_load(void);
...
+  const oc_endpoint_t
+    *endpoint; ///< endpoint to fetch the time (cannot be NULL)
+#if defined(OC_SECURITY) && defined(OC_PKI)
+  bool disable_time_verification; ///< ignore time validity checks when creating
+                                  ///< a new TLS session for plgd-time fetch
+  int selected_identity_cred_id;  ///< identity certificate to use for a created
+                                  ///< TLS session
+                                  ///< (-1 to use any available)
+#endif
+  uint16_t timeout; ///< timeout in seconds of the fetch time request (default
```
jkralik, line 197: what is the value of default timeout ?

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
   g_oc_plgd_time.set_system_time_data = set_system_time_data;
 }
 
+#ifdef OC_CLIENT
+
+plgd_time_fetch_config_t
+plgd_time_fetch_config(const char *uri, const oc_endpoint_t *endpoint,
```
jkralik, line 604: change the order - swap endpoint and uri.

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
...
+
+#if defined(OC_TCP) || defined(OC_SECURITY)
+
+static bool
+dev_time_is_connected(const oc_endpoint_t *endpoint)
+{
+#ifdef OC_SECURITY
+  if ((endpoint->flags & SECURED) != 0) {
+    OC_DBG("plgd-time: peer connected");
+    return oc_tls_connected(endpoint);
```
jkralik, line 748: use check if peer exist instead of `oc_tls_connected `
oc_tls_connected  - return true when handshake is over, but if handshake is in progress returns false, so you handle it as not connected  - so `oc_tls_add_peer` returns already exist peer.

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
...
+    return false;
+  }
+
+  if (disable_time_verification) {
+    OC_DBG("plgd-time: disable time verification for peer");
+    vcp->verify_certificate = peer->verify_certificate;
+    peer->user_data.data = vcp;
+    peer->user_data.free = time_verify_certificate_params_free;
+    peer->verify_certificate = dev_time_verify_certificate;
+  }
```
jkralik, line 733: leak `vcp` when `disable_time_verification` == false

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
...
+  oc_tls_peer_t *peer = oc_tls_add_peer(ep, MBEDTLS_SSL_IS_CLIENT);
+  if (peer == NULL) {
+    OC_ERR("plgd-time add peer failed: oc_tls_add_peer failed");
+    oc_memb_free(&g_time_verify_certificate_params_s, vcp);
+    return false;
+  }
+
+  if (disable_time_verification) {
+    OC_DBG("plgd-time: disable time verification for peer");
+    vcp->verify_certificate = peer->verify_certificate;
```
jkralik, line 729: maybe we need to also store free function clenaup previous user data.

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
...
+
+static int
+dev_time_verify_certificate(oc_tls_peer_t *peer, const mbedtls_x509_crt *crt,
+                            int depth, uint32_t *flags)
+{
+  OC_DBG("plgd-time: verifying certificate at depth %d, flags %u", depth,
+         *flags);
+
+  const time_verify_certificate_params_t *vcp =
+    (time_verify_certificate_params_t *)peer->user_data.data;
```
jkralik, line 695: set origin `user data` and `free` func to peer before this function is called.
`vcp->verify_certificate` can expect them

api/plgd/plgd_time.c
```diff
@@ -560,4 +598,243 @@ plgd_time_configure(bool use_in_mbedtls,
...
+  int ret = vcp->verify_certificate(peer, crt, depth, flags);
+  OC_DBG("plgd-time: default verification done (depth=%d, flags=%u)", depth,
```
jkralik, line 697: and set back `vcp` to user_data and free func

api/unittest/plgdtimetest.cpp
```diff
@@ -496,6 +501,120 @@ TEST_F(TestPlgdTimeWithServer, PutRequestFail)
...
+  if ((ep->flags & TCP) != 0) {
+    EXPECT_EQ(0,
+              oc_remove_session_event_callback_v1(tcp_events, nullptr, true));
+  }
+#endif /* OC_TCP && OC_SESSION_EVENTS */
+}
+
+TEST_F(TestPlgdTimeWithServer, FetchTimeAlreadyConnected)
+{
+  // TODO: use already connected endpoint
```
jkralik, line 613: Just this TODO :D

jkralik (approved):

PR #415, opened 2023-04-04
session events: update API

Danielius1922:


include/oc_session_events.h
```diff
@@ -16,65 +16,65 @@
...
 
-/**
-  @brief Callback function to pass the session event infomation to App.
-  @param endpoint  endpoint info which the session event is happened.
-  @param state  enum values in oc_session_state_t.
-*/
-typedef void (*session_event_handler_t)(const oc_endpoint_t *endpoint,
-                                        oc_session_state_t state);
+#define OC_ERR_SESSION_EVENT_HANDLER_NOT_FOUND -6
+#define OC_ERR_SESSION_EVENT_HANDLER_REMOVED -7
```
jkralik, line 44: This `OC_ERR_SESSION_EVENT_HANDLER_NOT_FOUND` never comes from the handler.

And OC_ERR_SESSION_EVENT_HANDLER_REMOVED is used only when networks is shutdown - but it is expected that all managers removed all registered callback. I think we don't need this.

include/oc_session_events.h
```diff
@@ -16,65 +16,65 @@
...
+ *              if < 0 an error value
+ * @param user_data user data provided by the user to
+ * oc_add_session_event_callback_v1
  *
- * @param endpoint stop event on endpoint
+ * @see oc_add_session_event_callback_v1
  */
-void oc_session_end_event(const oc_endpoint_t *endpoint);
+typedef void (*session_event_handler_v1_t)(const oc_endpoint_t *endpoint,
+                                           int state, void *user_data);
```
jkralik, line 70: if we more err states than state could be `oc_session_state_t` type

api/oc_session_events.c
```diff
@@ -162,6 +163,167 @@ oc_session_end_event(const oc_endpoint_t *endpoint)
...
+  }
+  if (session_event_remove_callback(oc_session_event_versioned_handler(cb),
+                                    NULL, false) < 0) {
+    return -1;
+  }
+  return 0;
+}
+
+int
+oc_remove_session_event_callback_v1(session_event_handler_v1_t cb,
```
jkralik, line 287: return bool (true for success, false for not exist)

include/oc_network_monitor.h
```diff
@@ -21,45 +21,84 @@
...
-*/
+ * @brief Add the callback to receive session event notifications.
+ *
+ * @param cb The callback to be added (cannot be NULL).
+ * @param user_data user data passed to the callback when invoked
+ * @return 0 on success
+ * @return -1 on error
+ */
+OC_API
+int oc_add_session_event_callback_v1(session_event_handler_v1_t cb,
```
jkralik, line 73: return bool instead of int ?

include/oc_network_monitor.h
```diff
@@ -21,45 +21,84 @@
...
+ oc_add_session_event_callback_v1
+ * @param ignore_user_data ignore user_data and match only the function pointer
+ * @return 0 on success
+ * @return OC_ERR_SESSION_EVENT_HANDLER_NOT_FOUND when no match was found
+ * @return -1 other errors
+ *
+ * @see oc_add_session_event_callback_v1
+ */
+OC_API
+int oc_remove_session_event_callback_v1(session_event_handler_v1_t cb,
```
jkralik, line 99: return `bool` instead of `int` ?

include/oc_network_monitor.h
Danielius1922: wanted to keep it consistent with `oc_add_session_event_callback` and `oc_remove_session_event_callback` which return error value.

jkralik (approved):

PR #416, opened 2023-04-05
PKI/TPM: allow to set pk functions for TPM use

jkralik:
To integrate TPM with iotivity-lite, the mbedtls library is used for cryptography. There are four main functions that need to be implemented:

1. **Generate public and private key**: The `mbedtls_pk_ecp_gen_key` function generates a private key for the CSR of the identity certificate when iotivity-lite doesn't have a private key in the keypair storage. In this function, a private key needs to be generated in the TPM, and the public key `mbedtls_pk_context` needs to be filled.
2. **Free private key in TPM**: The `pk_free_key` function frees the private key in the TPM. This function is called when a factory reset is performed or when the keypair storage does not match the TPM during the generation of the CSR with the reference key.
3. **Store reference key to credential resource**: The `mbedtls_pk_write_key_der` function writes the private key to a buffer for the credential resource. A reference key associated with the private key in the TPM needs to be created according to `mbedtls_pk_context *pk`. This reference key is written to the buffer and stored in the credential resource or keypair storage
4. **Load public key with reference key**: This function parses the public key/certificate and private key from the keypair storage or credential resource. The reference key to the private key in the TPM, as written by `mbedtls_pk_write_key_der`, needs to be used as the key to set the public key mbedtls_pk_context *pk by the TPM chip.

To make use of these TPM functions in iotivity-lite, you need to set the TPM implementations as `oc_pki_set_pk_functions` before calling `oc_main_init`.

```c

#include <iotivity-lite/oc_pki.h>

oc_pki_pk_functions_t pk_functions = {
    .mbedtls_pk_ecp_gen_key = tpm_mbedtls_pk_ecp_gen_key,
    .pk_free_key = simulate_tpm_pk_free_key,
    .mbedtls_pk_parse_key = tpm_mbedtls_pk_parse_key,
    .mbedtls_pk_write_key_der = tpm_mbedtls_pk_write_key_der,
};
if (!oc_pki_set_pk_functions(&pk_functions)) {
  // Failed to set PKI function
  return -1;
}
```

### DIFF
--- a/apps/cloud_server.c
+++ b/apps/cloud_server.c
@@ -186,6 +186,12 @@ static const char *cis;
 static const char *auth_code;
 static const char *sid;
 static const char *apn;
+#ifdef OC_PKI
+#include <mbedtls/sha256.h>
+static bool simulate_tpm = false;
+static uint8_t manufacturer_private_key[4096];
+const char *manufacturer_reference_private_key = "IDevID";
+#endif /* OC_PKI */
 #else  /* OC_SECURITY */
 static const char *cis = "coap+tcp://127.0.0.1:5683";
 static const char *auth_code = "test";
@@ -734,17 +740,28 @@ factory_presets_cb(size_t device, void *data)
   }
 
   unsigned char mfg_crt[4096];
-  size_t mfg_crt_len = 4096;
+  size_t mfg_crt_len = sizeof(mfg_crt);
   if (read_pem("pki_certs/mfgcrt.pem", (char *)mfg_crt, &mfg_crt_len) < 0) {
     PRINT("ERROR: unable to read pki_certs/mfgcrt.pem\n");
     return;
   }
   unsigned char mfg_key[4096];
-  size_t mfg_key_len = 4096;
+  size_t mfg_key_len = sizeof(mfg_key) - 1;
   if (read_pem("pki_certs/mfgkey.pem", (char *)mfg_key, &mfg_key_len) < 0) {
     PRINT("ERROR: unable to read pki_certs/mfgkey.pem\n");
     return;
   }
+  if (simulate_tpm) {
+    // set the manufacturer private key to the internal storage
+    memcpy(manufacturer_private_key, mfg_key, mfg_key_len);
+    manufacturer_private_key[mfg_key_len] = '\0';
+    // set reference private key as mfg_key
+    memcpy(mfg_key, manufacturer_reference_private_key,
+           strlen(manufacturer_reference_private_key));
+    mfg_key[strlen(manufacturer_reference_private_key)] = 0;
+    mfg_key_len = strlen(manufacturer_reference_private_key);
+  }
+
   int mfg_credid =
     oc_pki_add_mfg_cert(0, (const unsigned char *)mfg_crt, mfg_crt_len,
                         (const unsigned char *)mfg_key, mfg_key_len);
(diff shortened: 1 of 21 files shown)


security/unittest/csrtest.cpp
```diff
@@ -165,6 +165,145 @@ TEST_F(TestCSRWithDevice, Validate256)
...
+  static size_t public_key_size;
+  static std::array<uint8_t, OC_ECDSA_PRIVKEY_SIZE> private_key;
+  static size_t private_key_size;
+
+  static int simulate_tpm_mbedtls_pk_parse_key(
+    size_t device, mbedtls_pk_context *pk, const unsigned char *key,
+    size_t keylen, const unsigned char *pwd, size_t pwdlen,
+    int (*f_rng)(void *, unsigned char *, size_t), void *p_rng)
+  {
+    (void)device;
```
Danielius1922, line 204: in C++ you can delete the parameter name

security/oc_pki_internal.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <stddef.h>
+#include <mbedtls/ecp.h>
+#include <mbedtls/pk.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int oc_mbedtls_pk_parse_key(size_t device, mbedtls_pk_context *pk,
```
Danielius1922, line 30: Ask chatGPT to write documentation here :D

apps/cloud_server.c
```diff
@@ -738,11 +744,22 @@ factory_presets_cb(size_t device, void *data)
     return;
   }
   unsigned char mfg_key[4096];
-  size_t mfg_key_len = 4096;
```
Danielius1922, line 741: sizeof(mfg_key) - 1

security/oc_pki.c
```diff
@@ -415,4 +419,71 @@ oc_pki_get_verify_certificate_cb(void)
...
+    return mbedtls_ecp_gen_key(grp_id, mbedtls_pk_ec(*pk), f_rng, p_rng);
+  }
+  if (g_pk_functions->mbedtls_pk_ecp_gen_key == NULL) {
+    return MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE;
+  }
+  return g_pk_functions->mbedtls_pk_ecp_gen_key(device, grp_id, pk, f_rng,
+                                                p_rng);
+}
+
+OC_API
```
Danielius1922, line 475: `OC_API` should only be in the header

security/oc_pki.c
```diff
@@ -23,14 +23,17 @@
...
 #include "oc_store.h"
 #include "oc_tls.h"
 #include "port/oc_connectivity.h"
 
+#include <mbedtls/ecp.h>
 #include <mbedtls/pk.h>
 #include <mbedtls/x509_crt.h>
 
 static oc_pki_verify_certificate_cb_t g_verify_certificate_cb;
+static const oc_pki_pk_functions_t *g_pk_functions;
```
Danielius1922, line 36: Storing a pointer is dangerous if somebody sends a pointer to local variables to `oc_pki_set_pk_functions`. 

I would prefer `oc_pki_pk_functions_t` + `bool` for safety.

security/unittest/csrtest.cpp
Danielius1922, line 204: the `device` parameter is never used for `mbedtls_pk_parse_key_cb_t` nor for `mbedtls_pk_ecp_gen_key_cb_t`, why do we think it's necessary? wouldn't perhaps a `void*` argument for custom user data be more useful?

apps/cloud_server.c
jkralik: fixed

security/unittest/csrtest.cpp
jkralik: fixed

security/oc_pki.c
jkralik: removed

security/unittest/csrtest.cpp
jkralik: `void*` is not necessary, because pk functions are global, the `device` is used to determine which device asks for the certificate, in case you have multiple devices.

port/oc_log_internal.h
```diff
@@ -268,9 +270,15 @@
   do {                                                                         \
     char *beg = (buff);                                                        \
     char *end = (buff) + (size);                                               \
-    for (size_t i = 0; beg <= (end - 3) && i < (size_t)(len); i++) {           \
-      beg += (i == 0) ? SPRINTF(beg, "%02x", (data)[i])                        \
-                      : SPRINTF(beg, ":%02x", (data)[i]);                      \
+    /* without _oc_written = 9 has sometimes */                                \
```
Danielius1922, line 273: `_oc_written`? is that renamed `oc_log_ret`?

jkralik: yes - fixed

include/oc_pki.h
```diff
@@ -162,14 +169,156 @@ typedef int (*oc_pki_verify_certificate_cb_t)(struct oc_tls_peer_t *peer,
...
+ * @return false when any of the functions is NULL.
+ */
+OC_API
+bool oc_pki_set_pk_functions(const oc_pki_pk_functions_t *pk_functions);
+
+/**
+ * Get the PK functions for the identity certificate or the manufacturer
+ * certificate.
+ * @param[out] pk_functions the PK functions
+ * @param[out] set true if the PK functions have been set, false otherwise
```
Danielius1922, line 317: I would use the return value instead of this second output param

Danielius1922 (approved):

security/oc_csr.c
```diff
@@ -30,23 +30,60 @@
...
+}
+
+int
+oc_sec_csr_generate(size_t device, mbedtls_md_type_t md, unsigned char *csr,
+                    size_t csr_size)
+{
+  assert(csr != NULL);
+
+  mbedtls_pk_context pk;
+  mbedtls_pk_init(&pk);
```
Danielius1922, line 73: leak of pk_context on lines 91 and 96

Danielius1922: (I'll fix it, once I make lunch!)

PR #420, opened 2023-04-13
PlgdTime: compilation updates and issues fixes

Danielius1922:
- add -Wdocumentation flag
- run static analysis with latest clang(v15) and clang-tidy(v15)
- fix reported issues


swig/swig_interfaces/oc_session_events.i
```diff
@@ -25,32 +25,6 @@
...
-void jni_session_end_event(oc_endpoint_t *endpoint)
-{
-#ifdef OC_TCP
-  oc_session_end_event(endpoint);
-#else /* OC_TCP */
-  OC_DBG("JNI: %s - Must build with OC_TCP defined to use this function.\n", __func__);
-#endif /* !OC_TCP */
-}
-%}
-
```
jkralik, line 53: Why these is removed ?

Danielius1922: only public functions should be in the JAVA wrapper, these are internal

jkralik (approved):

PR #423, opened 2023-04-20
oc_clock_time_monotonic: use monotonic time from the boot time

jkralik:


port/esp32/adapter/src/clock.c
```diff
@@ -43,12 +45,8 @@ oc_clock_time(void)
 oc_clock_time_t
 oc_clock_time_monotonic(void)
 {
-  struct timespec t;
-  if (clock_gettime(CLOCK_MONOTONIC, &t) == -1) {
-    return -1;
-  }
-  return (oc_clock_time_t)t.tv_sec * OC_CLOCK_SECOND +
-         (oc_clock_time_t)ceil((double)t.tv_nsec / (1.e09 / OC_CLOCK_SECOND));
+  // return esp_timer_get_time returns microseconds from the boot time
```
Danielius1922, line 48: `OC_CLOCK_SECOND` might not be microseconds, so you need to do conversion to oc_clock_time_t

port/windows/clock.c
```diff
@@ -60,27 +54,9 @@ oc_clock_time(void)
...
-  // is sufficiently large.
-  const LONGLONG whole =
-    (qtime.QuadPart / g_query_perm_frequency) * OC_CLOCK_SECOND;
-  const LONGLONG part =
-    (LONGLONG)((qtime.QuadPart % g_query_perm_frequency) *
-               ((double)OC_CLOCK_SECOND / g_query_perm_frequency));
-  return whole + part;
+  // The return value is the number of milliseconds that have elapsed since the
+  // system was started.
+  return ((oc_clock_time_t)GetTickCount64()) * 1000;
```
Danielius1922, line 59: Why not use the more precise `QueryPerformanceCounter`. Also you need to do conversion to oc_clock_time_t

jkralik: Note that the QueryPerformanceCounter() function is hardware-dependent and may not be available on all systems. However, it is generally available on modern Windows systems and provides a high-resolution timer that is suitable for many timing applications.

port/esp32/adapter/src/clock.c
jkralik: fixed

jkralik:
~format

PR #431, opened 2023-04-26
logs: fix stack overflow when OC_LOGbytes is defined

jkralik:


python/oc_python.c
```diff
@@ -20,6 +20,11 @@
 
 // #define OC_SERVER
 
+#if defined(_WIN32)
```
Danielius1922, line 23: can be removed since you cast to int?

apps/simpleclient_windows.c
```diff
@@ -20,6 +20,7 @@
 #include "port/oc_clock.h"
 #include <signal.h>
 #include <windows.h>
+#include <inttypes.h>
```
Danielius1922, line 23: don't need this include anymore since you cast to int

python/oc_python.c
jkralik: removed

port/oc_log_internal.h
```diff
@@ -79,8 +80,24 @@
...
+#define PRINT_LOG(...) printf(__VA_ARGS__)
+#endif /* !PRINT_LOG */
+
 #ifndef PRINT
-#define PRINT(...) printf(__VA_ARGS__)
+#define PRINT(...)                                                             \
+  do {                                                                         \
+    oc_logger_t *logger = oc_log_get_logger();                                 \
+    if (logger->fn != NULL) {                                                  \
+      if (logger->level < (OC_LOG_LEVEL_DEBUG)) {                              \
```
Danielius1922, line 92: Hmm, this might breaks apps binaries or even CTT tool (it might read something from stdout). For example cloud_server will now print usage only if you compile it with debug level logs enabled, which doesn't seem right to me. 

Either we need to switch everything from apps, onboarding_tool and python to PRINT_LOG. Or we remove the usage of PRINT in library code - I've tried to do this in https://github.com/iotivity/iotivity-lite/pull/432 .

jkralik: Ok. I merged you changes to my PR

Danielius1922 (approved):

PR #435, opened 2023-04-28
provides API to set callback for oc_send_response

jkralik:
In cases of errors, iotivity-lite fails to populate the body. With the implementation of this callback function, you can customize and populate/replace the body as per your requirements.

### DIFF
--- a/api/oc_ri.c
+++ b/api/oc_ri.c
@@ -551,7 +551,7 @@ oc_ri_init(void)
 }
 
 static const char *method_strs[] = {
-  "EMPTY",  /* 0 */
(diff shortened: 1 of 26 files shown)


api/oc_ri.c
```diff
@@ -549,6 +550,23 @@ oc_ri_init(void)
   start_processes();
 }
 
+static const char *method_strs[] = {
+  "EMPTY",  /* 0 */
```
Danielius1922, line 554: We'll save 5 bytes if we use "" instead of "EMPTY" :D

Danielius1922 (approved):

PR #446, opened 2023-05-09
security:  when a connection has been closed, drop the messages associated with it

### LINKED ISSUE #441: TLS: closing of TLS peer vs. closing of TCP session
The IoTivity-lite TLS session consists of the two layers: a TCP session (takes care of the socket belonging to the TLS session) and a TLS peer (takes care of security). 
When a TLS session is closed then first the TLS peer is closed and deallocated (on the main thread) and message is sent to the network thread to stop and deallocate the corresponding TCP session. 

Unfortunately, there is a unhandled race condition when the TLS has been already deallocate, but the message to stop the TCP session hasn't been handled yet. It is possible for the TCP socket in this time window to receive a message and add it to the list of inbound messages waiting to be processed. The situation is illustrated in the log from IoTivity v2.2.5.4:

```text
main thread:
[OC 2023-05-01T19:48:27.180635Z] DEBUG: oc_tls.c:335 <oc_tls_free_peer>: oc_tls: freeing peer(0x55dc17485790)
[OC 2023-05-01T19:48:27.180645Z] DEBUG: oc_pstat.c:109 <print_pstat_dos>: oc_pstat: dos is RFNOP
[OC 2023-05-01T19:48:27.180654Z] DEBUG: observe.c:313 <coap_remove_observer_by_client>: Unregistering observers for client at: 
[OC 2023-05-01T19:48:27.180668Z] DEBUG: observe.c:314 <coap_remove_observer_by_client>: endpoint coaps+tcp://127.0.0.1:40030
[OC 2023-05-01T19:48:27.180747Z] DEBUG: observe.c:324 <coap_remove_observer_by_client>: Removed 0 observers
[OC 2023-05-01T19:48:27.180756Z] DEBUG: transactions.c:278 <coap_free_transactions_by_endpoint>: coap_free_transactions_by_endpoint

network thread:
[OC 2023-05-01T19:48:27.180863Z] DEBUG: ipadapter.c:1129 <process_events>: processing 1 events
[OC 2023-05-01T19:48:27.180880Z] DEBUG: oc_buffer.c:88 <allocate_message>: buffer: allocated message(0x7fdaf0004bd0) from pool(0x7fdaf78c88a0)
[OC 2023-05-01T19:48:27.180917Z] DEBUG: tcpsession.c:468 <tcp_receive_message>: tcp receive session(fd=20)
[OC 2023-05-01T19:48:27.180930Z] DEBUG: tcpsession.c:409 <tcp_session_receive_message>: recv(): 6 bytes.
[OC 2023-05-01T19:48:27.180939Z] DEBUG: oc_tcp.c:96 <oc_tcp_is_valid_header>: TLS header: record type: 21, major 3, minor 3
[OC 2023-05-01T19:48:27.180949Z] DEBUG: tcpsession.c:435 <tcp_session_receive_message>: tcp packet total length : 31 bytes.
[OC 2023-05-01T19:48:27.180960Z] DEBUG: tcpsession.c:409 <tcp_session_receive_message>: recv(): 25 bytes.
[OC 2023-05-01T19:48:27.180969Z] DEBUG: ipadapter.c:1051 <process_socket_read_event>: Incoming message of size 31 bytes from
[OC 2023-05-01T19:48:27.180979Z] DEBUG: ipadapter.c:1052 <process_socket_read_event>: endpoint coaps+tcp://127.0.0.1:40030
[OC 2023-05-01T19:48:27.180987Z] DEBUG: ipadapter.c:1053 <process_socket_read_event>: 
[OC 2023-05-01T19:48:27.180996Z] DEBUG: oc_network_events.c:133 <oc_network_receive_event>: inbound message(0x7fdaf0004bd0) for endpoint
[OC 2023-05-01T19:48:27.181061Z] DEBUG: tcpsession.c:493 <log_tcp_session>: found ongoing TCP session for
[OC 2023-05-01T19:48:27.181076Z] DEBUG: tcpsession.c:494 <log_tcp_session>: endpoint coaps+tcp://127.0.0.1:40030
[OC 2023-05-01T19:48:27.181085Z] DEBUG: tcpsession.c:495 <log_tcp_session>: 
[OC 2023-05-01T19:48:27.181109Z] DEBUG: ipadapter.c:1129 <process_events>: processing 1 events

main thread:
[OC 2023-05-01T19:48:27.181125Z] DEBUG: tcpsession.c:742 <free_session_async_locked>: signaled network event thread to monitor that the session needs to be removed
[OC 2023-05-01T19:48:27.181135Z] DEBUG: tcpsession.c:743 <free_session_async_locked>: free TCP session async

network thread:
[OC 2023-05-01T19:48:27.181200Z] DEBUG: tcpsession.c:290 <free_session_locked>: free TCP session(0x7fdaf0004b50, fd=20)
[OC 2023-05-01T19:48:27.181217Z] DEBUG: ipadapter.c:1013 <process_socket_signal_event>: Signal event received(fd=16, status=2)
```

The waiting inbound message is then processed by `oc_tls_recv_message` which checks if a TLS peer associated with the endpoint of the message exists, if it doesn't then it creates it.
At this moment we have a TLS peer that we wanted to close.

Proposed solution:
- add a flag to oc_endpoint_t for TCP (and DTLS?) connection to distinguish between server and client endpoints (examine the ACCEPTED flag if it can't be used)
- when an inbound message is processed by `oc_tls_recv_message` check if the receiving endpoint is flagged as a client and it that case new peer should't be opened. The message should be dropped and the event should be logged.
- examine if a similar issue cannot occur for an outbound message
- add unit tests

jkralik:
Fixes #441

After the TCP-TLS/DLTS session is closed, there may be queued messages in the TLS queue that are still being processed. If this occurs, the system will attempt to create a new connection. However, this can fail if the TLS peer is not properly configured, such as having the wrong role (server instead of client) or not allowing the appropriate ciphers for cloud connections.

Therefore, it is necessary to discard all messages associated with the events queue when the session ends.


util/oc_process.c
```diff
@@ -366,6 +366,38 @@ oc_process_is_closing_all_tls_sessions(void)
...
+    if (event->p != p || !drop_event(event->ev, event->data, user_data)) {
+      // move to the next event
+      ++i;
+      continue;
+    }
+    // we have a match, drop the event and move the last event to its position
+    if (nevents > 1) {
+      oc_process_num_events_t last_index =
+        (fevent + nevents - 1) % OC_PROCESS_NUMEVENTS;
+      events[index] = events[last_index];
```
Danielius1922, line 393: I am wondering whether changing the order of messages is dangerous or not.  From reading the headers I see no mention of a guaranteed order of execution. What do you think, won't we get some weird bugs?

security/oc_tls.c
```diff
@@ -329,10 +329,41 @@ oc_tls_free_invalid_peer(oc_tls_peer_t *peer)
...
+  if (ev != oc_events[RI_TO_TLS_EVENT] && ev != oc_events[UDP_TO_TLS_EVENT]) {
+    return false;
+  }
+  oc_message_t *message = (oc_message_t *)data;
+  if (oc_endpoint_compare(&message->endpoint, endpoint) == 0) {
+#if OC_DBG_IS_ENABLED
+    oc_string_t endpoint_str;
+    oc_endpoint_to_string(&message->endpoint, &endpoint_str);
+    OC_DBG("oc_tls: dropping %s message for removed endpoint(%s)",
+           (ev == oc_events[RI_TO_TLS_EVENT]) ? "sended" : "received",
```
Danielius1922, line 346: sended -> sent

api/oc_events.h
```diff
@@ -29,7 +29,7 @@ extern "C" {
 typedef enum {
   INBOUND_NETWORK_EVENT,
   UDP_TO_TLS_EVENT,
-  INIT_TLS_CONN_EVENT,
+  INIT_TLS_CONN_EVENT, /* deprecated - use RI_TO_TLS_EVENT */
```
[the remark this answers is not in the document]
jkralik: removed

util/oc_process.h
```diff
@@ -369,6 +369,30 @@ void oc_process_start(struct oc_process *p, oc_process_data_t data);
...
+ * be removed from the event queue of a given process.
+ *
+ * @param ev The event to be dropped.
+ * @param data The auxiliary data to be sent with the event
+ * @param user_data Data to be passed to the drop_event function.
+ * @return true Drop the event.
+ */
+typedef bool (*oc_process_drop_event_t)(oc_process_event_t ev,
+                                        oc_process_data_t data,
+                                        void *user_data);
```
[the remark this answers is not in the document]
jkralik: fixed

security/oc_tls.c
jkralik: fixed

security/oc_tls.c
```diff
@@ -418,6 +452,9 @@ oc_tls_remove_peer(const oc_endpoint_t *endpoint)
   oc_tls_peer_t *peer = oc_tls_get_peer(endpoint);
   if (peer != NULL) {
     oc_tls_free_peer(peer, false);
+  } else {
+    oc_process_drop(&oc_tls_handler, process_drop_event_for_removed_endpoint,
+                    (void *)endpoint);
```
Danielius1922, line 457: remove the cast

security/oc_tls.c
```diff
@@ -2449,6 +2494,18 @@ oc_tls_init_connection(oc_message_t *message)
 }
 #endif /* OC_CLIENT */
 
+size_t
+oc_tls_send_message(oc_message_t *message)
+{
+#if defined(OC_CLIENT)
+  if (!oc_tls_connected(&message->endpoint)) {
+    oc_tls_init_connection(message);
```
Danielius1922, line 2502: What if we get multiple calls to `oc_tls_send_message` with a peer that hasn't finished the TLS handshake yet. We will have multiple calls to `oc_tls_init_connection` -> multiple calls to `oc_tls_handshake` -> multiple calls to `mbedtls_ssl_handshake` with the same ssl context. Is that OK for mbedTLS?

jkralik: This is same behavior as was before. But as I known I expect that mbedtls has own state machine to control such thing, but I added check that the handshake is called just once.

security/oc_tls.c
jkralik: fixed

util/oc_process.c
jkralik: This is a issue for TCP connection - stream can be broken.

Danielius1922: I see, oh well. I guess we have to move the memory and keep the order - should be  0-2 memset calls (might be annoying to get the bounds and size right for the circular buffer) or a for-loop which moves all the elements forward by one.

jkralik: Fixed via memmove :)

Danielius1922: Looks good, tested it on paper with one case and the bounds were correct.

Danielius1922 (approved):

jkralik:
~format

security/oc_tls.c
```diff
@@ -390,12 +403,39 @@ oc_tls_free_peer(oc_tls_peer_t *peer, bool inactivity_cb)
...
   }
+  oc_process_drop(&oc_tls_handler, process_drop_event_for_removed_endpoint,
+                  &peer->endpoint);
 #ifdef OC_PKI
   oc_free_string(&peer->public_key);
 #endif /* OC_PKI */
   mbedtls_ssl_config_free(&peer->ssl_conf);
   oc_etimer_stop(&peer->timer.fin_timer);
+
+  oc_endpoint_t endpoint;
```
Danielius1922, line 414: Why is the local copy needed? It gets deallocated when the function ends, same way the peer::endpoint was previously. What is different?

jkralik: I moved deallocation before the call callbacks. Peer is a big struct so in constrain device you want to free resource. Eg in one of the callbacks there could be created new peer

PR #447, opened 2023-05-11
Linux: configure listening ports of a device

Danielius1922:
This repository provides a configuration option to specify port numbers or disable ports for each IP protocol. The following are the effects of disallowing specific protocols:

- Disabling UDP4 or UDP: The device will no longer be discoverable via multicast for certain protocol, and the connection will not work as both rely on the same socket.
- Disabling DTLS4 or DTLS: DTLS will be disabled for both the server and the client.
- Disabling TCP4 or TCP or TLS(TCP)4 or TLS(TCP): Only the TCP listener will be disabled, while the client side will continue to function.

Example: apps/cloud_server.c

### DIFF
--- a/port/linux/socklistener.h
+++ b/port/linux/socklistener.h
@@ -1,0 +1,69 @@
+/****************************************************************************
+ *
+ * Copyright (c) 2023 plgd.dev s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"),
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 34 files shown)


port/linux/socklistener.h
```diff
@@ -0,0 +1,53 @@
...
+#include <sys/socket.h>
+
+typedef struct oc_sock_listener_t
+{
+  struct sockaddr_storage sockaddr;
+  int sock;
+  uint16_t port;
+} oc_sock_listener_t;
+
+#define OC_CLOSE_SOCK_LISTENER(server)                                         \
```
Danielius1922, line 32: Why the use of macros instead of functions?

clang-tidy complains about stuff inside, see: 
https://github.com/iotivity/iotivity-lite/actions/runs/4980635641/jobs/8913797112

look for port/linux/socklistener.h

jkralik: fixed

jkralik (approved):

PR #450, opened 2023-05-16
tls: use monotonic timer for monitoring DTLS peer inactivity

Danielius1922:


security/oc_tls.c
```diff
@@ -529,25 +555,26 @@ oc_tls_handler_schedule_write(oc_tls_peer_t *peer)
...
+  }
+  oc_clock_time_t time = oc_clock_time_monotonic();
+  time -= peer->timestamp;
+  if (time < DTLS_INACTIVITY_TIMEOUT_TICKS) {
+    OC_DBG("oc_tls: Resetting DTLS inactivity callback");
+    return OC_EVENT_CONTINUE;
+  }
+  mbedtls_ssl_close_notify(&peer->ssl_ctx);
+  if ((peer->endpoint.flags & TCP) == 0) {
     mbedtls_ssl_close_notify(&peer->ssl_ctx);
```
jkralik, line 574: From this code it seems to be more general - not only for DTLS (udp tls), but also for TLS (TCP) connection

security/oc_tls.c
```diff
@@ -1994,8 +2025,8 @@ oc_tls_add_new_peer(oc_tls_new_peer_params_t params)
   if ((params.endpoint->flags & TCP) == 0) {
     mbedtls_ssl_set_timer_cb(&peer->ssl_ctx, &peer->timer, ssl_set_timer,
                              ssl_get_timer);
-    oc_ri_add_timed_event_callback_seconds(
-      peer, oc_tls_inactive, (oc_clock_time_t)OC_DTLS_INACTIVITY_TIMEOUT);
+    oc_ri_add_timed_event_callback_ticks(peer, oc_dtls_inactive,
+                                         DTLS_INACTIVITY_TIMEOUT_TICKS);
```
jkralik, line 2029: now I see the limitation to DTLS

jkralik, line 2029: Maybe this timer could be used to close also TCP-TLS connection when we implement the closing idle sessions for TCP

api/oc_endpoint.c
```diff
@@ -176,16 +121,25 @@ oc_endpoint_to_string(const oc_endpoint_t *endpoint, oc_string_t *endpoint_str)
     return -1;
   }
 
+  char ip[OC_IPV6_MINSTRLEN] = { 0 };
```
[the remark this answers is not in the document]
Danielius1922: Yeah, unclear variable name -> it was intented as minimal-required-size-to-fit-all-ipv6-address, which is equivalent to the size of the maximal ipv6 address. I'll rename it.

security/oc_tls.c
Danielius1922: I'm checking the header for `mbedtls_ssl_set_timer_cb` and it should work for TLS as well. It is mandatory for DTLS, but not for TLS so that's probably the reason why it is like this.

jkralik (approved):

PR #456, opened 2023-05-30
oc_introspection: disable feature in compile time

Danielius1922:
Add OC_INTROSPECTION_ENABLED switch to CMake, which controls whether the oc_introspection feature with /oc/wk/introspection and /oc/introspection resources is available.
(If disabled the code is not compiled in the library at all.)


Danielius1922:
@SiMet hello, could you please restart the CTT machine, something seems to be wrong. Thanks!

include/oc_rep.h
```diff
@@ -435,6 +435,18 @@ CborError oc_rep_encoder_close_container(CborEncoder *encoder,
     }                                                                          \
   } while (0)
 
+/** Alternative to oc_rep_set_text_string in case we know the length of the
+value */
+#define oc_rep_set_text_string_v1(object, key, value, value_len)               \
```
jkralik, line 440: pls in `oc_rep_set_text_string` call `oc_rep_set_text_string_v1`

Danielius1922: wanted to avoid doing a nil-check twice

jkralik (approved):

PR #474, opened 2023-06-26
storage: fix coverity issues

Danielius1922:
Issue 321044: Out-of-bounds write

### DIFF
--- a/messaging/coap/observe.c
+++ b/messaging/coap/observe.c
@@ -247,6 +247,8 @@ add_observer(oc_resource_t *resource, uint16_t block2_size,
 /*---------------------------------------------------------------------------*/
 /*- Removal -----------------------------------------------------------------*/
 /*---------------------------------------------------------------------------*/
+
+// TODO: use oc_string_view_t
 static const char *
 get_iface_query(oc_interface_mask_t iface_mask)
 {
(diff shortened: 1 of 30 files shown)


messaging/coap/observe.c
```diff
@@ -433,172 +435,196 @@ coap_remove_observer_by_resource(const oc_resource_t *rsc)
...
+#ifdef OC_BLOCK_WISE
+  uint16_t block2_size = obs->block2_size;
+#else  /* !OC_BLOCK_WISE */
+  uint16_t block2_size = 0;
+#endif /* OC_BLOCK_WISE */
+  if (coap_separate_accept(&req, response->separate_response, &obs->endpoint,
+                           obs->obs_counter, block2_size)) {
+    response->separate_response->active = 1;
+  }
+}
```
jkralik, line 468: This call is valid only for UDP/DTLS messages for TCP it is not supported - look to `coap_separate_accept` it is do nothing for TCP at the end ...

messaging/coap/observe.c
```diff
@@ -433,172 +435,196 @@ coap_remove_observer_by_resource(const oc_resource_t *rsc)
...
+
+#ifdef OC_BLOCK_WISE
+
+static int
+send_notification_blockwise(coap_packet_t *notification, coap_observer_t *obs,
```
jkralik, line 473: send_notification_blockwise -> prepare_notification_blockwise

it doesn't send anything

messaging/coap/observe.c
```diff
@@ -433,172 +435,196 @@ coap_remove_observer_by_resource(const oc_resource_t *rsc)
...
+  }
+  if (response->response_buffer->content_format > 0) {
+    coap_set_header_content_format(&notification,
+                                   response->response_buffer->content_format);
+  }
+  coap_set_token(&notification, obs->token, obs->token_len);
+  transaction = coap_new_transaction(coap_get_mid(), obs->token, obs->token_len,
+                                     &obs->endpoint);
+  if (transaction == NULL) {
+    return 0;
```
jkralik, line 616: free notification is not needed ?

Danielius1922: seems to be a stack object, doesn't allocate anything, so free/clear is not necessary

messaging/coap/observe.c
Danielius1922: Are you sure? Seems to me that `coap_separate_accept` stores token, uri and observe value to `response->separate_response`, which then can be send by `oc_send_separate_response`.

Issue #359, opened 2023-01-20
[ASAN] heap-use-after-free in post_cloud


Danielius1922:
Address sanitizer detects heap use after free when offboarding device from a cloud:

```
==507091==ERROR: AddressSanitizer: heap-use-after-free on address 0x625000223c0e at pc 0x7f3913f5d2c3 bp 0x7ffe711c0b30 sp 0x7ffe711c02d8
WRITE of size 1 at 0x625000223c0e thread T0
    #0 0x7f3913f5d2c2 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:827
    #1 0x5587176c31da in append_to_buffer /home/danielius/GIT/iotivity-lite/deps/tinycbor/src/cborencoder.c:297
    #2 0x5587176c32ad in append_byte_to_buffer /home/danielius/GIT/iotivity-lite/deps/tinycbor/src/cborencoder.c:305
    #3 0x5587176c3cae in create_container /home/danielius/GIT/iotivity-lite/deps/tinycbor/src/cborencoder.c:507
    #4 0x5587176c3d9a in cbor_encoder_create_map /home/danielius/GIT/iotivity-lite/deps/tinycbor/src/cborencoder.c:558
    #5 0x558717621edc in oc_rep_encoder_create_map_internal /home/danielius/GIT/iotivity-lite/api/oc_rep_encode.c:229
    #6 0x558717622009 in oc_rep_encoder_create_map /home/danielius/GIT/iotivity-lite/api/oc_rep_encode.c:247
    #7 0x5587176babc6 in cloud_response /home/danielius/GIT/iotivity-lite/api/cloud/oc_cloud_resource.c:59
    #8 0x5587176bbfb7 in post_cloud /home/danielius/GIT/iotivity-lite/api/cloud/oc_cloud_resource.c:177
    #9 0x558717629b87 in oc_ri_invoke_coap_entity_handler /home/danielius/GIT/iotivity-lite/api/oc_ri.c:1265
    #10 0x55871764631e in coap_receive /home/danielius/GIT/iotivity-lite/messaging/coap/engine.c:576
    #11 0x5587176490f2 in process_thread_g_coap_engine /home/danielius/GIT/iotivity-lite/messaging/coap/engine.c:937
    #12 0x5587176c19ae in call_process /home/danielius/GIT/iotivity-lite/util/oc_process.c:201
[...]

PR #477, opened 2023-06-29
android: disable network interface monitoring on ANDROID_API >= 30

Danielius1922:


port/android/ipadapter.c
```diff
@@ -66,6 +66,10 @@
 #include <sys/un.h>
 #include <unistd.h>
 
+#if __ANDROID_API__ < 30
```
baabel, line 69: should this be...  #if __ANDROID_API__ > 29

baabel:
Not necessarily related to this change but....

api/cloud/oc_cloud.c

when OC_NETWORK_MONITOR is not defined, the code in oc_cloud.c calls some undefined functions...(ex: oc_remove_network_interface_event_callback) and they end up as UNDEFINED in the .so library and loadLibrary() fails to load libiotivity-lite-jni.so.  This is because the implementation is conditional in ipadapter.c based on OC_NETWORK_MONITOR.

potential fix:
include the OC_NETWORK_MONITOR directive around calls line #342 and #359
example:
#if defined(OC_SESSION_EVENTS) && defined(OC_NETWORK_MONITOR)
  oc_remove_session_event_callback_v1(cloud_ep_session_event_handler, ctx,
                                      false);
  oc_remove_network_interface_event_callback(cloud_interface_event_handler);
  oc_add_session_event_callback_v1(cloud_ep_session_event_handler, ctx);
  oc_add_network_interface_event_callback(cloud_interface_event_handler);
#endif /* OC_SESSION_EVENTS */

port/android/oc_config.h
```diff
@@ -28,8 +28,13 @@ typedef uint64_t oc_clock_time_t;
 /* Maximum wait time for select function */
 #define SELECT_TIMEOUT_SEC (1)
 
+/* The current implementation in Android port relies on notifications from a
+ * NETLINK socket to retrieve interface change notifications, but this is
+ * restricted from ANDROID_API >= 30  */
+#if __ANDROID_API__ < 30
```
baabel, line 34: Should this be defined only for > 29 ?

Danielius1922:
> 

Yeah, the ifdefs are wrong, seems like `OC_SESSION_EVENTS` and `OC_NETWORK_MONITOR` got mixed up, I'll clean it up.

baabel:
@Danielius1922  Hi, I tested latest change. Looks like only 1 more undefined symbol remaining, so when launching android app ...dlopen failed: cannot locate symbol "oc_resource_set_access_in_RFOTM" referenced by...

Also ran nm on the .so library and that seems to be the only undefined symbol in iotivity lib

Danielius1922:
@baabel Should be all good now!

baabel:
thanks.  lgtm, android app launches.

Danielius1922:
Great, I've fired up the Conformance tests, if they all pass I'll merge the fixes to master (they usually take from 3 to 4 hours...)

PR #496, opened 2023-08-17
tls: print logs when a connection will been closed for error

jkralik:


security/oc_certs.c
```diff
@@ -396,10 +396,12 @@ oc_certs_parse_CN_buffer_for_UUID(mbedtls_asn1_buf val, char *buffer,
...
-#if OC_ERR_IS_ENABLED
+#if OC_DBG_IS_ENABLED
     oc_string_t cn;
     oc_new_string(&cn, uuid_CN, val.len);
-    OC_ERR("invalid Common Name field (tag:%d val:%s)", val.tag, oc_string(cn));
+    OC_DBG("Common Name field (tag:%d val:%s) is not in format " UUID_PREFIX
+           ":<UUID string>",
+           val.tag, oc_string(cn));
     oc_free_string(&cn);
 #endif /* OC_ERR_IS_ENABLED */
```
Danielius1922, line 406: OC_DBG_IS_ENABLED

security/oc_tls.c
```diff
@@ -69,13 +69,50 @@
 #include <mbedtls/debug.h>
 #include <mbedtls/error.h>
 #include <mbedtls/platform.h>
-#endif /* OC_DEBUG */
+#elif OC_DBG_IS_ENABLED || OC_ERR_IS_ENABLED || OC_WRN_IS_ENABLED
+static void
+mbedtls_strerror(int ret, char *buf, size_t buflen)
```
Danielius1922, line 74: I'm surprised SC doesn't complain about includes after a function/variable definition.  I guess we have logs disabled for the analysis build so it doesn't skewer the results, but I would move this to line 85

security/oc_tls.c
```diff
@@ -69,13 +69,50 @@
...
+#endif /* OC_DBG_IS_ENABLED || OC_ERR_IS_ENABLED || OC_WRN_IS_ENABLED */
 
 #include <stdarg.h>
 #include <stdint.h>
 #include <string.h>
 #include <inttypes.h>
 
+#if OC_DBG_IS_ENABLED || OC_ERR_IS_ENABLED
+#if OC_ERR_IS_ENABLED
+#define MBEDTLS_ERR(mbedtls_func_name, mbedtls_err)                            \
```
Danielius1922, line 88: Are preprocessor functions instead of normal ones necessary?

security/oc_tls.c
```diff
@@ -1368,6 +1431,11 @@ static int
 oc_tls_load_identity_cert_chain(mbedtls_ssl_config *conf, size_t device,
                                 int credid)
 {
+  if (credid < -1) {
```
Danielius1922, line 1434: for `oc_tls_load_identity_cert_chain` is -1 OK?

security/oc_tls.c
jkralik: For log line in proper places - yes

security/oc_tls.c
jkralik: moved

security/oc_tls.c
jkralik: added define for ANY. Yes, any is a default behavior.

security/oc_certs.c
jkralik: fixed

security/oc_tls.c
```diff
@@ -1368,6 +1435,11 @@ static int
 oc_tls_load_identity_cert_chain(mbedtls_ssl_config *conf, size_t device,
                                 int credid)
 {
+  if (credid < OC_TLS_SELECTED_ANY_CRED_ID) {
```
Danielius1922, line 1438: I see, we use this feature in DPS, where we use `-2` to skip identity certificates and force mfg certificates. We should document it in some relevant header, because it might get removed if it is not mentioned anywhere.

security/oc_tls.c
Danielius1922: Yeah, line will be different, though you have the `mbedtls_func_name` label which can be used to uniquely identify the log.

Danielius1922 (approved):

security/oc_tls.c
```diff
@@ -1342,32 +1379,74 @@ oc_tls_configure_end_entity_cert_chain(mbedtls_ssl_config *conf, size_t device,
...
 {
   OC_DBG("loading manufacturer cert chain");
   return oc_tls_configure_end_entity_cert_chain(conf, device,
                                                 OC_CREDUSAGE_MFG_CERT, credid);
 }
 
-static int
-oc_tls_load_identity_cert_chain(mbedtls_ssl_config *conf, size_t device,
-                                int credid)
+#ifndef OC_TEST
```
Danielius1922, line 1438: Just put the declaration outside the `OC_TEST` ifdef in the header. `oc_tls_validate_identity_certs_consistency` and `oc_tls_validate_trust_anchors_consistency` declarations are ifdef-ed in the header because the definitions exist only with `OC_TEST`, otherwise internal functions have declarations in internal headers to be available for tests.

jkralik: Fixed

PR #508, opened 2023-08-28
Batch observation of /oic/res with ETag

Danielius1922:


messaging/coap/observe.c
```diff
@@ -1262,47 +1271,104 @@ notify_batch_observer(coap_observer_t *obs, oc_response_t *response)
...
   response_buffer.buffer = buffer;
   response_buffer.buffer_size = OC_MIN_OBSERVE_SIZE;
+
+  // TODO: cache of etags for batch interface to avoid repeated calculation?
+  // struct {
+  //   const oc_resource_t* resource;
+  //   const oc_endpoint_t* endpoint;
+  //   uint64_t etag;
+  //}
+  // oc_list_t etags;
```
jkralik, line 1354: For each observation etags could be different.

messaging/coap/observe.c
```diff
@@ -1262,47 +1271,104 @@ notify_batch_observer(coap_observer_t *obs, oc_response_t *response)
...
     const oc_blockwise_state_t *response_state =
       oc_blockwise_find_response_buffer(oc_string(obs->resource->uri) + 1,
                                         oc_string_len(obs->resource->uri) - 1,
                                         &obs->endpoint, OC_GET, query.data,
                                         query.length, OC_BLOCKWISE_SERVER);
     if (response_state != NULL) {
+      COAP_DBG(
+        "response_state is not NULL, sending of batch response currently "
+        "in progress for endpoint, skipping to next observer");
+      // TODO: reschedule?
```
jkralik, line 1371: By [RFC ](https://datatracker.ietf.org/doc/html/rfc7959#section-2.4) the current response need to be dropped and new notification need to be send. But with this code the reschedule is planned because it is not removed from the list (line 1392) - we are safe :)

jkralik (approved):

messaging/coap/observe.c
Danielius1922: Yes, but if nothing happens the notification might be stuck in the list for a long time, I was thinking maybe about scheduling a dispatch in a minute or something.

jkralik: When blockwise transfer finishes then rescheduling is expected via cbk in blockwise.

messaging/coap/observe.c
jkralik: For same endpoint with same href and method no.

jkralik (approved):

PR #527, opened 2023-09-13
Check iotivity-lite stack init before processing requests

jkralik:
check if the iotivity-lite stack is initialized before processing new requests during shutdown.

```
#0  0x00007f5052dfa213 in ...
#1  0x00007f505306325f in ...
#2  0x00007f50527d1935 in oc_core_device_handler () from target:/lib/libiotivity-lite-client-server.so.2
#3  0x00007f50527e0fb1 in oc_ri_invoke_coap_entity_handler () from target:/lib/libiotivity-lite-client-server.so.2
#4  0x00007f50527eb799 in coap_receive () from target:/lib/libiotivity-lite-client-server.so.2
#5  0x00007f50527ec5b1 in process_thread_g_coap_engine () from target:/lib/libiotivity-lite-client-server.so.2
#6  0x00007f50527c8bb4 in call_process () from target:/lib/libiotivity-lite-client-server.so.2
#7  0x00007f50527c8f16 in do_event () from target:/lib/libiotivity-lite-client-server.so.2
#8  0x00007f50527c8f36 in oc_process_run () from target:/lib/libiotivity-lite-client-server.so.2
#9  0x00007f50527d8c96 in oc_main_poll () from target:/lib/libiotivity-lite-client-server.so.2
#10 0x00007f50527e26dc in oc_ri_shutdown () from target:/lib/libiotivity-lite-client-server.so.2
#11 0x00007f50527d8cc7 in oc_main_shutdown () from target:/lib/libiotivity-lite-client-server.so.2
```

### DIFF
--- a/messaging/coap/engine.c
+++ b/messaging/coap/engine.c
@@ -521,6 +521,11 @@ coap_receive_method_payload(coap_receive_ctx_t *ctx, const char *href,
   }
 #endif /* OC_SECURITY */
 
+  if (!oc_main_initialized()) {
+    COAP_DBG("cannot process new requests during shutdown iotivity-lite stack");
+    return COAP_RECEIVE_ERROR;
+  }
+
 #ifdef OC_TCP
   bool is_valid_size =
     ((endpoint->flags & TCP) != 0 &&
--- a/messaging/coap/engine.c
+++ b/messaging/coap/engine.c
@@ -488,11 +488,9 @@ coap_receive_blockwise_block2(coap_receive_ctx_t *ctx, const char *href,
 }
 
 static coap_receive_status_t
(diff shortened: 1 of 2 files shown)


messaging/coap/engine.c
```diff
@@ -521,6 +521,11 @@ coap_receive_method_payload(coap_receive_ctx_t *ctx, const char *href,
   }
 #endif /* OC_SECURITY */
 
+  if (!oc_main_initialized()) {
```
Danielius1922, line 524: This seems a bit late, can't we put this at the beginning of `coap_process_inbound_message` to skip the message receive at the start and avoid executing most of the code?

Danielius1922: We decided to keep it like this, maybe the check can be executed sooner, but that will require more extensive testing.

PR #530, opened 2023-09-14
use oc_endpoint_to_string64 to utilize stack

### LINKED ISSUE #529: Optimize memory allocation based on heaptrack results
[heaptrack.txt](https://github.com/iotivity/iotivity-lite/files/12600146/heaptrack.txt)

jkralik:
Fixes #529

### DIFF
--- a/api/oc_endpoint.c
+++ b/api/oc_endpoint.c
@@ -16,8 +16,9 @@
  *
  ****************************************************************************/
 
+#include "api/oc_endpoint_internal.h"
+#include "api/oc_helpers_internal.h"
 #include "oc_endpoint.h"
-#include "oc_endpoint_internal.h"
 #include "oc_core_res.h"
 #include "port/common/oc_ip.h"
 #include "port/oc_connectivity.h"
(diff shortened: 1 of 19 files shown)


include/oc_helpers.h
```diff
@@ -36,6 +37,13 @@
 extern "C" {
 #endif
 
+typedef struct oc_string64_s
+{
+  struct oc_string64_s *next;
```
Danielius1922, line 42: Do you have some future use in mind where `oc_string64_t` is put into an `oc_list_t`? From the uses in this PR we don't need that, so the `next` member is not necessary, in fact we skip the whole type and just use `char[64]`.

api/oc_endpoint.c
```diff
@@ -66,26 +66,50 @@ oc_endpoint_set_di(oc_endpoint_t *endpoint, const oc_uuid_t *di)
   memcpy(endpoint->di.id, di->id, sizeof(di->id));
 }
 
-const char *
-oc_endpoint_flags_to_scheme(unsigned flags)
+bool
+oc_endpoint_flags_to_scheme(unsigned flags, char *buf, size_t *buf_len)
```
Danielius1922, line 70: I'd rather avoid in-out parameters (`buf_len` in this case) if we can help it. I would copy `snprintf` -> take the size of the buffer in arguments, return the number of written bytes or -1 on error.

include/oc_helpers.h
jkralik: removed next + moved to internal header.

api/oc_endpoint.c
jkralik: fixed

include/oc_helpers.h
```diff
@@ -24,6 +24,7 @@
 #include "oc_export.h"
 #include "util/oc_compiler.h"
 #include "util/oc_list.h"
+#include "util/oc_macros_internal.h"
```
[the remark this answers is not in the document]
jkralik: removed

PR #540, opened 2023-10-03
ETag Resource Checksum

Danielius1922:


util/unittest/crc64test.cpp
```diff
@@ -0,0 +1,51 @@
...
+    uint64_t crc;
+  };
+
+  std::vector<test> inputs{
+    { "", 0x0 },
+    { "Hello, World!", 0xA885B0FA12A6B582 },
+    { "1234567890", 0x4CCE99FD976EC1A8 },
+    { "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+      0x9A076C1F9CFD879C },
+    { "IoTivity-Lite is the best IoT C library on the flat plane!",
```
jkralik, line 41: 😄

api/oc_server_api.c
```diff
@@ -142,7 +142,11 @@ oc_send_response_internal(oc_request_t *request, oc_status_t response_code,
   }
   request->response->response_buffer->content_format = content_format;
   request->response->response_buffer->response_length = response_length;
-  request->response->response_buffer->code = oc_status_code(response_code);
+  if (response_code < __NUM_OC_STATUS_CODES__) {
+    request->response->response_buffer->code = oc_status_code(response_code);
+  } else {
+    request->response->response_buffer->code = response_code;
+  }
```
jkralik, line 149: this seems to be weird

api/oc_rep_encode_crc.c
```diff
@@ -0,0 +1,156 @@
...
+    .encode_boolean = &rep_crc_encode_bool,
+    .encode_int = &rep_crc_encode_int,
+    .encode_uint = &rep_crc_encode_uint,
+    .encode_floating_point = &rep_crc_encode_floating_point,
+    .encode_double = rep_crc_encode_double,
+    .encode_byte_string = &rep_crc_encode_byte_string,
+    .encode_text_string = NULL,
+    .create_array = NULL,
+    .create_map = NULL,
+    .close_container = NULL,
```
jkralik, line 152: What happens when they are not set ? It seems they throw segfault. eg:
```
CborError err = g_rep_encoder.close_container(encoder, containerEncoder);
```
Could you implement them similar as `encode_null` ?

api/oc_rep_encode_crc_internal.h
```diff
@@ -0,0 +1,46 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "api/oc_rep_encode_internal.h"
+
+#define OC_CRC_REP_NULL (0x0)
+
+#define OC_CRC_REP_TRUE (0x1)
+#define OC_CRC_REP_FALSE (0x0)
```
jkralik, line 35: this value need to be differ than `OC_CRC_REP_NULL` -> for each special type.

api/oc_rep_encode_crc.c
Danielius1922: Yes, I will this PR is not finished, I just wanted to get some results from SC.

api/oc_rep_encode_crc_internal.h
Danielius1922: Do we need a unique representations though? You have many values of different times that will have the same CRC result:

For example doubles and int64_t with the same bit pattern:

0xABCDEFGH is equal to some integer, some double and some uint8_t* byte array / string, all these different values map to the same bit pattern and therefore will have the same CRC value.

Maybe we should add one byte representing the type of the value to each CRC calculation?

api/oc_rep_encode.c
```diff
@@ -181,100 +222,117 @@ oc_rep_encoder_set_type_by_accept(oc_content_format_t accept)
   return false;
 }
 
-oc_content_format_t
+int
```
jkralik, line 225: Why int ? If you want to signal error change definintion to:
`bool oc_rep_encoder_get_content_format(ct* oc_content_format_t)`

include/oc_ri.h
```diff
@@ -505,7 +508,7 @@ const char *oc_status_to_str(oc_status_t key);
  * @brief Convert method to string. It is thread safe.
  *
  * @return Method in const char *.
- * @return Empty string for an invalid log level value
+ * @return Empty string for an method value
```
jkralik, line 511: invalid method

api/oc_etag_internal.h
```diff
@@ -120,6 +120,18 @@ bool oc_etag_load_from_storage(bool from_storage_only);
...
+/** Do not dump resource with given URI to storage. */
+bool oc_etag_dump_ignore_resource(const char *uri, size_t uri_len) OC_NONNULL();
+
+enum {
+  OC_RESOURCE_CRC64_OK, ///< resource has a payload and crc64 is calculated
+  OC_RESOURCE_CRC64_NO_PAYLOAD, ///< resource has no payload
+};
+
+/** Calculate crc64 checksum for given resource */
+int oc_resource_get_crc64(oc_resource_t *resource, uint64_t *crc64)
```
jkralik, line 132: return `enum` instead of `int`

jkralik (approved):

jkralik:
Good job 👍

jkralik:
@Danielius1922 Pls could you integrate load and dump etag to cloud_server ?

PR #565, opened 2023-11-09
Set interface index from incoming message if not set in TLS peer

jkralik:
This commit handles the case where the connection is initialized by IoTivity-Lite (oc_tls_init_connection).

Example: Cloud Manager

### DIFF
--- a/security/oc_tls.c
+++ b/security/oc_tls.c
@@ -3020,6 +3020,13 @@ oc_tls_recv_message(oc_message_t *message)
          u, message->length, (void *)peer);
 #endif /* OC_DBG_IS_ENABLED */
 
+  // Set the interface index from the incoming message if the network interface
+  // is not set in the peer. This case occurs when the connection is initialized
+  // by IoTivity-Lite (oc_tls_init_connection). Example: Cloud Manager
+  if (peer->endpoint.interface_index == 0) {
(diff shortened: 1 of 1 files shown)


security/oc_tls.c
```diff
@@ -3020,6 +3020,13 @@ oc_tls_recv_message(oc_message_t *message)
          u, message->length, (void *)peer);
 #endif /* OC_DBG_IS_ENABLED */
 
+  // Set the interface index from the incoming message if the network interface
+  // is not set in the peer. This case occurs when the connection is initialized
+  // by IoTivity-Lite (oc_tls_init_connection). Example: Cloud Manager
+  if (peer->endpoint.interface_index == 0) {
+    peer->endpoint.interface_index = message->endpoint.interface_index;
```
Danielius1922, line 3027: This also adds interface index to endpoints that were not initialized by `oc_tls_init_connection`.  Could that be a problem? I guess not, those should have the interface index set from the beginning if I recall correctly.

Danielius1922 (approved):

security/oc_tls.c
jkralik: No. As you said, the interface could be set from the beginning.
When you create TLS connection you don't know which interface would be used (ofcourse you can set it), so you leave it with 0 and when it receive the message than you know which interface has been used.

PR #569, opened 2023-11-24
Optimize resource usage by reducing the size of oc_message_t buffer

jkralik:


messaging/coap/coap.c
```diff
@@ -1021,17 +1021,7 @@ coap_oscore_parse_options(coap_packet_t *packet, const uint8_t *data,
       ++current_option;
       packet->payload = current_option;
       packet->payload_len = (uint32_t)(data_len - (packet->payload - data));
-
-      if (packet->transport_type == COAP_TRANSPORT_UDP &&
-          packet->payload_len > (uint32_t)OC_MAX_APP_DATA_SIZE) {
-        packet->payload_len = (uint32_t)OC_MAX_APP_DATA_SIZE;
```
Danielius1922, line 1027: This was a weird code. I think removing it should be OK, because below we check if the options overflow the data buffer:

```C
if (current_option + option_length > data + data_len) {
```

so we should return `BAD_REQUEST_4_00` if we get a message that's too long here.

api/oc_message_internal.h
```diff
@@ -57,15 +57,34 @@ oc_message_t *oc_message_allocate_outgoing(void);
...
  * @brief Get size of the message buffer
  *
  * The size of the buffer depends on compilation options (it may be static or
  * dynamic and have differing sizes).
  *
+ * @param message the message to get the buffer size
  * @return size_t size of the message buffer
  */
-size_t oc_message_buffer_size(void);
+size_t oc_message_buffer_size(const oc_message_t *message);
```
Danielius1922, line 79: I would add `OC_NONNULL()` to `oc_message_buffer_size` and `oc_message_shrink_buffer` and remove the 

```C
  if (message == NULL) {
    ...
  }
```

checks. Tthese are internal functions which are never used with NULL, we don't have to worry about user inputs, so compiler checks are sufficient.

jkralik: Fixed

api/oc_message.c
```diff
@@ -205,3 +226,30 @@ oc_message_unref(oc_message_t *message)
   OC_DBG("buffer: freed TX/RX buffer; num free: %d", oc_memb_numfree(pool));
 #endif /* !OC_DYNAMIC_ALLOCATION || OC_INOUT_BUFFER_POOL */
 }
+
+void
+oc_message_shrink_buffer(oc_message_t *message, size_t size)
```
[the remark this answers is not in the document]
jkralik: Added

api/oc_message.c
```diff
@@ -44,29 +47,29 @@ message_deallocate(oc_message_t *message, struct oc_memb *pool)
 #if defined(OC_DYNAMIC_ALLOCATION) && !defined(OC_INOUT_BUFFER_SIZE)
```
Danielius1922, line 47: `defined(OC_DYNAMIC_ALLOCATION) && !defined(OC_INOUT_BUFFER_SIZE)` -> `OC_HAS_FEATURE_MESSAGE_SHRINK_BUFFER`

Seems that there are several places with the old defines. I would switch it to the new feature macro for consistency.

util/oc_features.h
```diff
@@ -67,4 +67,12 @@
 #define OC_HAS_FEATURE_ETAG_INTERFACE
 #endif /* OC_HAS_FEATURE_ETAG && OC_STORAGE */
 
+#if defined(OC_DYNAMIC_ALLOCATION) && !defined(OC_INOUT_BUFFER_SIZE)
+#define OC_HAS_FEATURE_MESSAGE_SHRINK_BUFFER
```
Danielius1922, line 71: I'm thinking `OC_HAS_FEATURE_MESSAGE_DYNAMIC_BUFFER` would be better if we use it to ifdef not only the shrink function.

Danielius1922 (approved):

api/oc_message.c
jkralik: Sure - fixed

util/oc_features.h
jkralik: Renamed

port/esp32/adapter/src/ipadapter.c
```diff
@@ -877,7 +878,7 @@ network_event_thread(void *data)
       OC_DBG("interface index %d", message->endpoint.interface_index);
       OC_DBG("%s", "");
 
-#if defined(OC_DYNAMIC_ALLOCATION) && !defined(OC_INOUT_BUFFER_SIZE)
+#ifdef OC_HAS_FEATURE_MESSAGE_DYNAMIC_BUFFER
       void *tmp = realloc(
         message->data,
         message->length +
```
Danielius1922, line 884: We can get rid of the +1, since we removed the write from `coap_oscore_parse_options`

jkralik: Removed because shrink is done by the func `oc_network_receive_event(message);`

PR #570, opened 2023-11-24
Fix issues reported by coverity

Danielius1922:
Fix "Bad comparison of floating-point expressions" when encoding tag_pos_rel.


jkralik (approved):

port/common/posix/oc_poll.c
```diff
@@ -0,0 +1,64 @@
...
+  if (ret < 0) {
+    return -1;
+  }
+  if (ret == 0) {
+    return 0;
+  }
+
+  for (nfds_t i = 0; i < nfds; ++i) {
+    if (fds[i].revents != 0 && !on_event(fds[i].fd, fds[i].revents, data)) {
+      break;
```
jkralik, line 43: why only one event is processed ?

jkralik (commented):
https://github.com/iotivity/iotivity-lite/pull/570/files#r1409861707

port/common/posix/oc_poll.c
```diff
@@ -0,0 +1,64 @@
...
+
+// make ppoll() available
+#define _GNU_SOURCE
+
+#include "port/common/posix/oc_poll_internal.h"
+#include "oc_clock_util.h"
+
+#include <unistd.h>
+
+static int
```
jkralik, line 29: `int` is wired when you processing events because `on_event` can stop processing so you don't process all n so maybe change it to enum like: `error,timeout,event` ?

PR #581, opened 2023-12-13
apps/cloud_server: enable a mocked software update resource

jkralik:


apps/cloud_server.c
```diff
@@ -1604,7 +1671,7 @@ main(int argc, char *argv[])
   oc_storage_config("./cloud_server_creds/");
 #endif /* OC_STORAGE */
   oc_set_factory_presets_cb(factory_presets_cb, NULL);
-  oc_set_max_app_data_size(8 * 1024 + num_resources * 512);
+  oc_set_max_app_data_size(8 * 1024 + (num_resources + 1) * 512);
```
Danielius1922, line 1674: if we want to save up some memory then +1 should be only necessary if `OC_SOFTWARE_UPDATE` is defined

.github/workflows/plgd-device-test-with-cfg.yml
```diff
@@ -43,7 +43,7 @@ on:
         required: false
         default: docker/apps/Dockerfile.cloud-server-debug
 env:
-  TEST_CLOUD_SERVER_IMAGE: ghcr.io/plgd-dev/device/test-cloud-server:latest
```
Danielius1922, line 46: revert before merge

Danielius1922 (approved):

PR #586, opened 2024-01-09
Bridge manager

joochlee:
- **bridge_manager**
	- This branch includes `Bridge Manager` which interworks with `Bridge CLI` and manages Ecosystem Translation plugin modules
	- most of sonarqube code smells were fixed in local environment
	- written in C
	- `readme.me` is included


joochlee:
will be request again after `merge bridging` is merged

PR #587, opened 2024-01-09
Matter translation plugin

joochlee:
- **matter_translation_plugin**
	- This branch includes `Matter Translation plugin module` which translates between OCF Device and CSA Matter Endpoint
	- most of sonarqube code smells were fixed in local environment
	- written in C\+\+
	- `readme.me` is included


joochlee:
will be request again after `merge bridging` is merged

PR #592, opened 2024-01-12
Merge bridging

joochlee:
merge_bridging : fix `size_t` formatter, doxygen keyword issues


api/oc_bridge.c
```diff
@@ -51,19 +51,19 @@ OC_LIST(g_vods);
 static oc_resource_t *g_vodlist_res;
 
 #define OC_PRINT_VODSLIST                                                      \
```
Danielius1922, line 53: This looks like a debug print to me so I would printf only if `OC_DBG_IS_ENABLED` holds.

Also `OC_DBG` prints source file lines, so I understood why you used `OC_PRINTVODSLIST` as a macro, but `OCPRINTF` doesn't, so you can use a function now

Something like:

```C
static void print_vodlist()
{
#if OC_DBG_IS_ENABLED
  OC_PRINTF("\"vods\": [");                                                       
  oc_vods_t *print_vod_item = (oc_vods_t *)oc_list_head(g_vods);       
  while (print_vod_item) {                                                     
    OC_PRINTF("  {");                                                             
    OC_PRINTF("    \"n\": \"%s\"", oc_string(print_vod_item->name));              
    char di_uuid[OC_UUID_LEN];                                                 
    oc_uuid_to_str(&print_vod_item->di, di_uuid, OC_UUID_LEN);                 
    OC_PRINTF("    \"di\": \"%s\"", di_uuid);                                     
    OC_PRINTF("    \"econame\": \"%s\"", oc_string(print_vod_item->econame));     
    if (print_vod_item->next) {                                                
      OC_PRINTF("  },");                                                          
    } else {                                                                   
      OC_PRINTF("  }");                                                           
    }                                                                          
    print_vod_item = print_vod_item->next;                                     
  }
#endif
}

```

joochlee: thanks, it is fixed.

PR #595, opened 2024-01-22
merge_bridging: revert some changes, add unittest for most of new API…

joochlee:
revert some changes, add unittest for most of new APIs, fix sonarCloud codesmells/security hot spots


joochlee:
Daniel, 

Some build/test failed. but I don't know why.
Could you check this?

PR #604, opened 2024-02-05
cloud: add support for multiple servers for a single cloud

Danielius1922:
Remove oc_cloud_store_t and oc_cloud_context_t definitions from oc_cloud.h header.

### DIFF
--- a/api/cloud/oc_cloud_endpoint.c
+++ b/api/cloud/oc_cloud_endpoint.c
@@ -34,8 +34,8 @@ cloud_endpoint_uri_is_valid(oc_string_view_t uri)
 }
 
 static void
-cloud_endpoints_set_selected(oc_cloud_endpoints_t *ce,
-                             const oc_cloud_endpoint_t *selected)
+cloud_endpoint_select(oc_cloud_endpoints_t *ce,
+                      const oc_cloud_endpoint_t *selected)
 {
   if (ce->selected == selected) {
     return;
@@ -44,6 +44,17 @@ cloud_endpoints_set_selected(oc_cloud_endpoints_t *ce,
   if (ce->on_selected_change != NULL) {
     ce->on_selected_change(ce->on_selected_change_data);
   }
+}
+
+bool
+oc_cloud_endpoint_select(oc_cloud_endpoints_t *ce,
+                         const oc_cloud_endpoint_t *selected)
+{
+  if (!oc_list_has_item(ce->endpoints, selected)) {
+    return false;
+  }
+  cloud_endpoint_select(ce, selected);
+  return true;
 }
 
 static oc_cloud_endpoint_t *
@@ -68,7 +79,7 @@ cloud_endpoint_item_allocate_and_add(oc_cloud_endpoints_t *ce,
   // automatically select the first endpoint added
   if (ce->selected == NULL) {
     assert(oc_list_length(ce->endpoints) == 0);
-    cloud_endpoints_set_selected(ce, cei);
+    cloud_endpoint_select(ce, cei);
   }
 
   oc_list_add(ce->endpoints, cei);
@@ -182,7 +193,7 @@ oc_cloud_endpoints_clear(oc_cloud_endpoints_t *ce)
     cloud_endpoint_item_free(cei);
     cei = oc_list_pop(ce->endpoints);
   }
-  cloud_endpoints_set_selected(ce, NULL);
+  cloud_endpoint_select(ce, NULL);
 }
 
 typedef struct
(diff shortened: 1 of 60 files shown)


api/cloud/oc_cloud_store_internal.h
```diff
@@ -28,6 +28,22 @@
 extern "C" {
 #endif
 
+typedef struct oc_cloud_store_t
+{
+  oc_string_t ci_server;     ///< URL of OCF Cloud.
+  oc_string_t auth_provider; ///< The name of the Authorisation Provider through
+                             // which access token was obtained.
+  oc_string_t uid;
```
Danielius1922, line 36: @jkralik any idea what is this `uid` property? It's not mentioned in specification, and it is only used by cloud resource GET/POST handlers, otherwise it's not used anywhere in code.

jkralik: It is needed for non-jwt token. It determines uid of cloud user that can manipulate with device.
It is specified in https://openconnectivity.org/specs/OCF_Cloud_Security_Specification.pdf page 15. (table 4)

api/cloud/oc_cloud_endpoint.c
```diff
@@ -0,0 +1,285 @@
...
+void
+oc_cloud_endpoints_iterate(const oc_cloud_endpoints_t *ce,
+                           oc_cloud_endpoints_iterate_fn_t fn, void *data)
+{
+  if (ce->endpoints == NULL) {
+    return;
+  }
+  oc_cloud_endpoint_t *cei = (oc_cloud_endpoint_t *)oc_list_head(ce->endpoints);
+  while (cei != NULL) {
+    if (!fn(cei, data)) {
```
jkralik, line 120: in `fn` user can delete `cei` so we need to guard this.

api/cloud/oc_cloud_manager.c
```diff
@@ -316,20 +317,27 @@ cloud_manager_handle_redirect_response(oc_cloud_context_t *ctx,
...
+      size == 0) {
+    return false;
+  }
+  oc_string_view_t cis = oc_string_view(value, size);
+  if (!oc_cloud_endpoint_contains(&ctx->store.ci_servers, cis) &&
+      oc_cloud_endpoint_add(&ctx->store.ci_servers, cis) == NULL) {
+    OC_ERR("failed to add server to the list");
+    return false;
+  }
+  if (oc_cloud_endpoint_is_selected(&ctx->store.ci_servers, cis)) {
```
jkralik, line 330: the origin `cis` need to be removed to leave backward compatibility.

jkralik:
Points from the meeting:
- The selected CIS need to be stored in the storage, so for every new selection, we need to dump it to storage.
- DPS will reprovision the cloud after 30 seconds (watcher). However, we need to ensure that the selected iteration has not been dropped if the list hasn't changed, and the cloud is still unable to reach the device (the new CIS will be selected in the worst case after 128 seconds).

api/cloud/oc_cloud_endpoint.c
Danielius1922: added a comment to documentation

api/oc_rep.c
```diff
@@ -146,80 +146,97 @@ oc_free_rep(oc_rep_t *rep)
...
+}
+
 static bool
 oc_rep_get_value(const oc_rep_t *rep, oc_rep_value_type_t type, const char *key,
                  size_t key_len, void **value, size_t *size)
 {
   if (rep == NULL) {
     OC_ERR("Error of input parameters: invalid rep");
-    return false;
+    return NULL;
```
Danielius1922, line 178: false

apps/cloud_server.c
```diff
@@ -1535,6 +1535,21 @@ cloud_server_log(oc_log_level_t log_level, oc_log_component_t component,
...
+  if (cis == NULL || auth_code == NULL || sid == NULL) {
+    return;
+  }
+  oc_cloud_provision_conf_resource(ctx, cis, auth_code, sid, apn);
+
+  oc_uuid_t id = {
+    { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 15, 16 },
+  };
+  oc_cloud_add_server(ctx, "coap+tcp://mock.cloud:5683",
+                      CHAR_ARRAY_LEN("coap+tcp://mock.cloud:5683"), id);
```
jkralik, line 1550: why this is added here ? - it only works for unsecure device.

jkralik (approved):

apps/cloud_server.c
Danielius1922: ah sorry, I was testing the GET output for device with multiple servers, needs to be removed

Danielius1922: removed

PR #655, opened 2025-05-04
Fix coverity issues

Danielius1922:


Danielius1922:
~format

Danielius1922:
~format

api/oc_server_api.c
```diff
@@ -617,19 +617,19 @@ handle_separate_response_request(coap_separate_t *request,
   bool blockwise = response_buffer->response_length > request->block2_size;
 #endif /* OC_TCP */
   if (blockwise) {
+    const char *uri = oc_string(request->uri);
+    size_t uri_len = oc_string_len(request->uri);
```
jkralik, line 621: In [api/oc_client_api.c](https://github.com/iotivity/iotivity-lite/pull/655/files#diff-9c9927de713c0d6be777c6a9d5d8566db754aa27198f7d7392dc344902646ac3) oc_string_len_unsafe is used. But here we use `oc_string_len` - why ?

Danielius1922: In oc_client_api.c the struct used is oc_client_cb_t, which can never have a null uri. There the nullcheck confused the analyzer, so it thought that the the pointer can possibly be null and thus len - 1 would overflow a give a large invalid value. Using the unsafe version should prevent the analyzer to consider this code branch.

Here `coap_separate_t` is used, and I'm not familiar with it, I think the uri value should also never be null, but there is no further calculation done here, so we can keep this version and the analyzer won't complain.

jkralik (approved):

